PR #3, opened 2021-07-07
chore(docker): reduce dockerfile resulting image size

laurentlp:
This PR reduces the docker image size by a considerable amount using multistage builds, only having production dependencies on the resulting image and using Alpine Linux as the base image.

|         | Base Image           | Size  |
| ------------- |:-------------| -----:|
| Before      | node:12.13.0 | 1.34GB |
| After      | node:12.13.0-alpine |  405MB |

Closes MES-2

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -9,3 +9,10 @@ bin/
 doc
 res
 scripts
+.editorconfig
+.eslintrc.js
+.gitignore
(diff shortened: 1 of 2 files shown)


samuelmasse (approved):
That's a pretty big difference good job

PR #7, opened 2021-07-08
feat(redis): added support to prefix redis channels

laurentlp:
This PR adds support for specifying a Redis scope (`REDIS_SCOPE`) which is then used to prefix Redis channels and keys as this prevents messages to be published across every subscribers on a Redis cluster (no matter what DB they subscribed on). See: https://redis.io/topics/pubsub#database-amp-scoping

Closes MES-4

### DIFF
--- a/res/config.json
+++ b/res/config.json
@@ -14,7 +14,8 @@
   "redis": {
     "enabled": false,
     "connection": null,
-    "options": null
+    "options": null,
+    "scope": null
   },
   "sync": [
     {
--- a/src/distributed/redis/config.ts
+++ b/src/distributed/redis/config.ts
@@ -4,4 +4,5 @@ export interface RedisConfig {
   enabled: boolean
   connection: string | ClusterNode[]
   options: RedisOptions | ClusterOptions
+  scope: string
 }
--- a/src/distributed/redis/subservice.ts
+++ b/src/distributed/redis/subservice.ts
@@ -13,6 +13,7 @@ export class RedisSubservice implements DistributedSubservice {
   private sub!: Redis
   private callbacks: { [channel: string]: (message: any) => Promise<void> } = {}
   private pings!: PingPong
+  private scope!: string
 
   constructor(private config: RedisConfig) {}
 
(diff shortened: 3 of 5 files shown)


src/distributed/redis/subservice.ts
```diff
@@ -13,13 +13,15 @@ export class RedisSubservice implements DistributedSubservice {
...
   private pings!: PingPong
+  private scope!: string
 
   constructor(private config: RedisConfig) {}
 
   async setup() {
     this.nodeId = Math.round(Math.random() * 1000000)
     this.logger.info(`Id is ${clc.bold(this.nodeId)}`)
 
+    this.scope = process.env.REDIS_SCOPE || this.config.scope
```
samuelmasse, line 24: You forgot to use the redis.scope setting in the .json file so I added this here

src/global.ts
```diff
@@ -9,6 +9,7 @@ export interface MessagingEnv {
   CLUSTER_ENABLED?: string
   REDIS_URL?: string
   REDIS_OPTIONS?: string
+  REDIS_SCOPE?: string
```
samuelmasse, line 12: Always good to add typings for env vars here.

src/distributed/redis/subservice.ts
```diff
@@ -87,11 +89,19 @@ export class RedisSubservice implements DistributedSubservice {
...
   }
 
   async send(channel: string, message: any) {
-    await this.pub.publish(channel, JSON.stringify({ nodeId: this.nodeId, ...message }))
+    const scopedChannel = this.makeScopedChannel(channel)
+
+    await this.pub.publish(scopedChannel, JSON.stringify({ nodeId: this.nodeId, ...message }))
+  }
+
+  private makeScopedChannel(key: string): string {
```
samuelmasse, line 104: I put the helper function at the bottom since it's private and not very useful to see before the other functions

src/distributed/redis/config.ts
```diff
@@ -4,4 +4,5 @@ export interface RedisConfig {
   enabled: boolean
   connection: string | ClusterNode[]
   options: RedisOptions | ClusterOptions
+  scope: string
```
samuelmasse, line 7: While most of the codebase has no typings for the config.json file, here it happens that the redis service uses typings sur I added the scope property here.

res/config.json
```diff
@@ -14,7 +14,8 @@
   "redis": {
     "enabled": false,
     "connection": null,
-    "options": null
+    "options": null,
+    "scope": null
```
samuelmasse, line 18: It's useful to add new properties to the config.json file. Even if set to null it helps to know that this setting exists.

samuelmasse (approved):
Good job :+1: . I made a few nitpick comments. Also you forgot to use the redis.scope setting from the json file to I fixed that

PR #15, opened 2021-07-13
chore(dev): uniform logging

allardy:
Closes MES-48

I've made some minor changes to log files just so they fit better with the surrounding logs.

Since we have nlu, studio and messaging, we need to prepend log messages to indicate which process printed that log. The port is "useless" for the end user so i've grayed it, and we don'T need to print the external url since the server already prints it

image

### DIFF
--- a/src/logger/types.ts
+++ b/src/logger/types.ts
@@ -49,7 +49,7 @@ export class Logger {
     const time = moment().format(timeFormat)
 
     const timeText = clc.blackBright(time)
-    const titleText = clc.bold(this.colors[level](this.scope))
+    const titleText = clc.bold(this.colors[level](`[Messaging] ${this.scope}`))
 
     if (data) {
       // eslint-disable-next-line no-console
--- a/src/logger/types.ts
+++ b/src/logger/types.ts
@@ -1,5 +1,6 @@
 import clc from 'cli-color'
 import moment from 'moment'
+import yn from 'yn'
 
 export enum LoggerLevel {
   Debug = 10,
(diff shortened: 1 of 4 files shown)


src/logger/types.ts
```diff
@@ -49,7 +49,7 @@ export class Logger {
     const time = moment().format(timeFormat)
 
     const timeText = clc.blackBright(time)
-    const titleText = clc.bold(this.colors[level](this.scope))
+    const titleText = clc.bold(this.colors[level](`[Messaging] ${this.scope}`))
```
samuelmasse, line 52: Having `[Messaging]` in front of every log is annoying if messaging is started seperatly. So it would be better to only have that if messaging was spinned from the main repo.

samuelmasse (changes requested):
I can see how this makes the logging better when messaging is spinned from the main repo. I don't really like the changes for when messaging is run seperatly though.

What do you think of having a `SPINNED` environment variables that decides between the two styles of logging? It can be set to true when the messaging server is spinned from the main repo.

Also I decided to use `clc` for this project instead of `chalk` so please avoid using it

allardy:
Is there really a use case where messaging will be executed as a standalone, without BP? I see one with NLU, but absolutely none for messaging... 

As for why I used chalk, I just copy-pasted the code from the main repository to print it the same way. Chalk is used by like 15 libraries in the project, so while it's not defined by itself in package.json, it still work  because it's used by them

samuelmasse (approved):
Closes MES-48

PR #31, opened 2021-07-20
fix(twilio): add support for old webhooks

laurentlp:
This PR adds support for old webhooks on Twilio by re-constructing the URL that was proxies.

Closes MES-64


samuelmasse (approved):
:+1:

PR #32, opened 2021-07-21
feat(messenger): improved error handling and logging

laurentlp:
This PR adds:

 - Handles the case where the provider exists by has no corresponding conduit (e.g. channel is disabled,  and not synced)
 - Only display the error message when a sender throw an error (since the error stack may contain Axios' response with sensitive information)
 -  Added typings and better logging on the `auth` function that handles payload validation

### DIFF
--- a/src/channels/base/conduit.ts
+++ b/src/channels/base/conduit.ts
@@ -64,7 +64,7 @@ export abstract class ConduitInstance<TConfig, TContext extends ChannelContext<a
         try {
           renderer.render(context)
         } catch (err) {
-          this.logger.error('An error occurred when rendering a message', err)
+          this.logger.error('An error occurred when rendering a message', (err as Error).message)
         } finally {
           context.handlers++
         }
--- a/src/channels/base/conduit.ts
+++ b/src/channels/base/conduit.ts
@@ -76,7 +76,7 @@ export abstract class ConduitInstance<TConfig, TContext extends ChannelContext<a
         try {
           await sender.send(context)
         } catch (err) {
-          this.logger.error('An error occurred when sending a message', err)
+          this.logger.error('An error occurred when sending a message:', (err as Error).message)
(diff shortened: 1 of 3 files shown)


src/channels/base/conduit.ts
```diff
@@ -76,7 +76,7 @@ export abstract class ConduitInstance<TConfig, TContext extends ChannelContext<a
         try {
           await sender.send(context)
         } catch (err) {
-          this.logger.error('An error occurred when sending a message', err)
+          this.logger.error('An error occurred when sending a message:', (err as Error).message)
```
samuelmasse, line 79: I figured if this worked here that I'd also add this for the renderer errors

src/channels/messenger/channel.ts
```diff
@@ -64,14 +65,16 @@ export class MessengerChannel extends Channel<MessengerConduit> {
...
 
-    const signature = req.headers['x-hub-signature']
+    const signature = req.headers['x-hub-signature'] as string
     const [, hash] = signature.split('=')
     const expectedHash = crypto.createHmac('sha1', conduit.config.appSecret).update(buffer).digest('hex')
     if (hash !== expectedHash) {
-      throw new Error("Couldn't validate the request signature.")
+      this.logger.error("Couldn't validate the request signature. Make sure your appSecret is properly configured")
+
+      throw new Error()
```
samuelmasse, line 77: Kind of weird. It seems the error is catched somewhere in the express.json middleware and printed to the console without the red color. This works alright as an alternative I guess.

samuelmasse (approved):
:+1:

src/channels/base/conduit.ts
laurentlp: The reason I did not do the same thing for the sender is because the renderer will never print sensitive information in its error stack

PR #33, opened 2021-07-21
feat(smooch): add error handling during webhook setup

laurentlp:
This PR adds error handling during the webhook setup. Otherwise, it would crash Messaging if the call fails!


samuelmasse (approved):
More generally we could also have log errors whenever a initialize() or setup() call on a conduit fails

laurentlp:
> More generally we could also have log errors whenever a initialize() or setup() call on a conduit fails

Yes totally agree, I will create an issue for this!!

PR #34, opened 2021-07-21
feat(teams): improved asyncMiddleware and typings

laurentlp:
Self-explanatory


src/channels/base/channel.ts
```diff
@@ -63,7 +63,10 @@ export abstract class Channel<TConduit extends ConduitInstance<any, any>> {
     return (req: Request, res: Response, next: NextFunction) => {
       fn(req, res, next).catch((e) => {
         this.logger.error(`Error occurred calling route ${req.originalUrl}:`, e)
-        return res.sendStatus(500)
+
+        if (!res.headersSent) {
```
laurentlp, line 67: Since `conduit.adapter.processActivity` sends the headers before failing, we have to make sure that the headers have not been sent before doing so.

samuelmasse (approved):
:+1:

PR #37, opened 2021-07-22
feat(telegram): improved card rendering

laurentlp:
This PR improves card rendering with Telegram.

#### Before:

Screenshot from 2021-07-22 10-27-32

#### After:

Screenshot from 2021-07-22 10-26-09

Related to https://github.com/botpress/botpress/pull/5204


samuelmasse (changes requested):
What happens if the card doesn't have any image? Is the message still sent?

laurentlp:
> What happens if the card doesn't have any image? Is the message still sent?

Right!!! Good catch, it's now fixed!

samuelmasse (approved):
:+1:

PR #48, opened 2021-07-23
feat(client): implement http client

samuelmasse:
This client is meant to be used has an npm package by the old repo (and any other services) to communicate with messaging. It provides typings for everything

I think a new package should be made to hold typings that are shared between the `client` and the `server` package. It could be called `shared` or `common` and would avoid copy pasting typings from the server to the client (like I did here for example).

PR in main repo https://github.com/botpress/botpress/pull/5209

Closes MES-60

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ node_modules/
 dist/
 bin/
 *.log
+.tsbuildinfo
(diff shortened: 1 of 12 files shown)


samuelmasse:
@laurentlp I'll merge this right away for convenience but I'd like your opinion on the implementation here

PR #51, opened 2021-07-23
fix(vonage): added support for file rendering + improved carousel and video renderers

laurentlp:
This PR adds:

 - .nvmrc file
 - File renderer
 - Captions to video renderer
 - Handling of undefined subtitle in carousel/card


packages/server/src/channels/vonage/renderers/video.ts
```diff
@@ -20,7 +20,9 @@ export class VonageVideoRenderer implements ChannelRenderer<VonageContext> {
         type: 'video',
         text: undefined!,
         video: {
-          url: formatUrl(context.botUrl, payload.video)!
+          url: formatUrl(context.botUrl, payload.video)!,
+          // @ts-ignore
+          caption: payload.title!
```
samuelmasse, line 25: Why is there a ts ignore here?

laurentlp: Because the typings that the Vonage SDK exposes are simply wrong or do not fit their own API!

samuelmasse: I'd prefer a cast to any over a ts-ignore in that case

samuelmasse (approved):
:+1:

PR #53, opened 2021-07-24
feat(vonage): added support for image, button, file, video and location reception

laurentlp:
This PR adds support for image, button, file, video and location reception with channel Vonage.


samuelmasse (approved):
:+1:

PR #54, opened 2021-07-26
feat(telegram): added webhook validation

laurentlp:
This PR adds a check to validate the webhook endpoint so that we can be sure all requests come from Telegram.

As per [Telegram's documentation](https://core.telegram.org/bots/api#setwebhook):

> If you'd like to make sure that the Webhook request comes from Telegram, we recommend using a secret path in the URL, e.g. https://www.example.com/<token>. Since nobody else knows your bot's token, you can be pretty sure it's us.

Closes MES-14

### DIFF
--- a/packages/server/src/channels/telegram/channel.ts
+++ b/packages/server/src/channels/telegram/channel.ts
@@ -1,3 +1,4 @@
+import { Request } from 'express'
 import { Channel } from '../base/channel'
 import { TelegramConduit } from './conduit'
 import { TelegramConfigSchema } from './config'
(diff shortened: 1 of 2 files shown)


packages/server/src/channels/telegram/channel.ts
```diff
@@ -25,16 +26,26 @@ export class TelegramChannel extends Channel<TelegramConduit> {
...
+        req.url = `/${req.params.token}`
 
         const conduit = res.locals.conduit as TelegramConduit
-        conduit.callback(req, res)
+        if (this.validate(conduit, req)) {
+          conduit.callback(req, res)
+        } else {
+          this.logger.error("Request validation failed. Request probably didn't come from Telegram.")
+
+          res.status(401).send('Auth token invalid')
```
samuelmasse, line 40: Seems like a lot of text that says the same thing. I would remove the logger.error from here and just do
```ts
return res.sendStatus(401)
```

packages/server/src/channels/telegram/channel.ts
```diff
@@ -25,16 +26,26 @@ export class TelegramChannel extends Channel<TelegramConduit> {
...
+        }
       })
     )
 
     this.printWebhook()
   }
+
+  private validate(conduit: TelegramConduit, req: Request): boolean {
+    return req.params.token === conduit.config.botToken
+  }
```
samuelmasse, line 50: It's kind of an unnecessary overhead to extract this as a function here. There original function is already tiny so putting that in there won't hurt readability.

samuelmasse (approved):
:+1: 

I left a few nitpick comments otherwise seems all good!

packages/server/src/channels/telegram/channel.ts
laurentlp: I totally agree, that in this case, the function is unnecessary, but I think it would be nice if all channels implemented a validate function so we can maybe factorize the code a little and, by default, add this function as middleware or something like that!

packages/server/src/channels/telegram/channel.ts
laurentlp: Good point!

PR #57, opened 2021-07-26
fix(channels): fix request proxying requiring to use router.use

laurentlp:
This PR fixes an issue where all routes In messaging add to be defined using router.use because of a misconfiguration of the middleware proxy in the main repo (see: https://github.com/botpress/botpress/pull/5112/commits/035a123b507a3e77369f5ea3631bde2eca5afc95)

Closes MES-41

### DIFF
--- a/packages/server/src/channels/twilio/channel.ts
+++ b/packages/server/src/channels/twilio/channel.ts
@@ -24,9 +24,10 @@ export class TwilioChannel extends Channel<TwilioConduit> {
   async setupRoutes() {
     this.router.use(express.urlencoded({ extended: true }))
 
-    this.router.use(
+    this.router.post(
       '/',
       this.asyncMiddleware(async (req, res) => {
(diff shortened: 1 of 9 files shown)


packages/server/src/channels/twilio/channel.ts
```diff
@@ -24,9 +24,10 @@ export class TwilioChannel extends Channel<TwilioConduit> {
   async setupRoutes() {
     this.router.use(express.urlencoded({ extended: true }))
 
-    this.router.use(
+    this.router.post(
       '/',
       this.asyncMiddleware(async (req, res) => {
+        console.dir(req)
```
samuelmasse, line 30: You forgot a console here

laurentlp: Woupsi! Good catch!

samuelmasse (approved):
:+1:

Issue #56, opened 2021-07-26
Ignore unknown properties when validating a payload


allardy:
I got this error message and had to spend some time to find the issue.

![image](https://user-images.githubusercontent.com/42552874/127041598-f55dace2-c2d8-4c16-a72d-6cd91af73626.png)

Somehow I had 2 different configurations:
![image](https://user-images.githubusercontent.com/42552874/127041720-a6417e79-8916-457b-9a0a-3031cfcca119.png)

When validating the schema using Joi, please consider setting the option "ignore unknown properties" instead of rejecting the payload completely. The correct fields were also present, so the bot would have worked properly.

samuelmasse:
`clientId` and `clientToken` was renamed to `id` and `token` at some commit somewhere. It crashed because you ran the server using data that was generated by some indev version and now you're using a newer indev version (and there's obviously no migration between the two).

allardy:
@samuelmasse I know that's the cause, I was just suggesting a better approach to be lenient when validating payloads. We care about properties we are expecting and that needs to be validated. Superfluous properties are better ignored than to be completely rejected.

@laurentlp thanks for the fix

PR #58, opened 2021-07-26
fix(sync): strip unknown properties when syncing

laurentlp:
This PR fixes https://github.com/botpress/messaging/issues/56


samuelmasse (approved):
:+1:

PR #62, opened 2021-07-27
fix(channels): loosen joi schema validation by discarding unknown properties

laurentlp:
This PR loosens Joi schema validation by discarding unknown properties instead of throwing an error.

### DIFF
--- a/packages/server/src/instances/service.ts
+++ b/packages/server/src/instances/service.ts
@@ -95,7 +95,7 @@ export class InstanceService extends Service {
     try {
       await instance.initialize()
     } catch (e) {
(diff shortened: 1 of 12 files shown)


packages/server/src/instances/service.ts
```diff
@@ -95,7 +95,7 @@ export class InstanceService extends Service {
     try {
       await instance.initialize()
     } catch (e) {
-      instance.logger.error('Error trying to initialize conduit.', e)
+      instance.logger.error('Error trying to initialize conduit.', (e as Error).message)
```
laurentlp, line 98: Otherwise, it prints telegram's webhook which contains the bot token!

PR #66, opened 2021-07-29
fix(dockerfile): adapt dockerfile to work with yarn workspaces

laurentlp:
Self-explanatory for the Dockerfile part and changes the status check response code to 200.


samuelmasse (approved):
:+1:

PR #67, opened 2021-07-29
feat(batching): implement batching

samuelmasse:
Implements batching for tables that have high rates of inserts

Closes MES-38


packages/server/src/conversations/service.ts
```diff
@@ -66,6 +91,8 @@ export class ConversationService extends Service {
   }
 
   public async getMostRecent(clientId: uuid, userId: uuid): Promise<Conversation | undefined> {
+    // TODO: need to figure out batching for this
```
samuelmasse, line 94: These functions are kind of wrong to be in the conversation service since they use messages. We'll move them eventually but for now they'll stay there. They kind of don't work with batching (because we'd need to flush the messages batch which we don't have access to). We'll leave it like that for now and label batching as experimental and we'll come back to this when we want to finalize the batching implementation

laurentlp (approved):

PR #71, opened 2021-08-02
chore(channels): use absolute urls

samuelmasse:
This PR removes the botUrl and formatUrl stuff from the codebase. We assume that media is sent using absolute urls

Closes MES-81


laurentlp (approved):
LGTM!

PR #72, opened 2021-08-02
feat(health): implement health service

samuelmasse:
Implements a health service that allows the api user to track the state of their configured channels.

Closes MES-82

### DIFF
--- a/packages/server/src/database/service.ts
+++ b/packages/server/src/database/service.ts
@@ -82,6 +82,10 @@ export class DatabaseService extends Service {
   }
 
   async destroy() {
+    // This delay is necessary to assure that all queries to the database are done before closing the connection
+    const delay = 1000
+    await new Promise((resolve) => setTimeout(resolve, delay))
+
     await this.knex.destroy()
   }
 
--- a/packages/server/src/health/service.ts
+++ b/packages/server/src/health/service.ts
@@ -1,22 +1,41 @@
 import { Service } from '../base/service'
+import { uuid } from '../base/types'
 import { ChannelService } from '../channels/service'
 import { ClientService } from '../clients/service'
 import { ConduitService } from '../conduits/service'
+import { DatabaseService } from '../database/service'
+import { InstanceService } from '../instances/service'
 import { ProviderService } from '../providers/service'
+import { HealthTable } from './table'
+import { HealthEventType } from './types'
+import { HealthWatcher } from './watcher'
 
 export class HealthService extends Service {
+  private table: HealthTable
+  private watcher: HealthWatcher
+
   constructor(
+    private db: DatabaseService,
     private channelService: ChannelService,
     private providerService: ProviderService,
     private clientService: ClientService,
-    private conduitService: ConduitService
+    private conduitService: ConduitService,
+    private instanceService: InstanceService
   ) {
     super()
+    this.table = new HealthTable()
+    this.watcher = new HealthWatcher(this.conduitService, this.instanceService, this)
   }
 
-  async setup() {}
+  async setup() {
+    await this.db.registerTable(this.table)
 
-  async getHealth(clientId: string) {
+    await this.watcher.setup()
+  }
+
+  async register(conduitId: uuid, type: HealthEventType, info: any = undefined) {}
+
+  async getHealth(clientId: uuid) {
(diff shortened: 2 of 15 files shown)


packages/server/src/database/service.ts
```diff
@@ -82,6 +82,10 @@ export class DatabaseService extends Service {
   }
 
   async destroy() {
+    // This delay is necessary to assure that all queries to the database are done before closing the connection
+    const delay = 1000
```
laurentlp, line 86: What guarantee do you have that after a second all queries will have finished?

packages/server/src/health/service.ts
```diff
@@ -0,0 +1,147 @@
...
+  }
+
+  // TODO: move this somewhere else
+  private async callWebhook(url: string, data: any) {
+    const password = process.env.INTERNAL_PASSWORD || this.configService.current.security?.password
+
+    try {
+      await axios.post(url, data, { headers: { password } })
+    } catch (e) {
+      // TODO: maybe we should retry if this call fails
```
laurentlp, line 90: Totally agree. We might have to look at something like an incremental backoff strategy (as most channels do).

packages/server/src/instances/events.ts
```diff
@@ -0,0 +1,21 @@
+import { Emitter } from '../base/events'
+import { uuid } from '../base/types'
+
+// Note: event type ids should be random numbers between -9007199254740991 and 9007199254740991
```
laurentlp, line 4: And why is that?

packages/server/src/instances/service.ts
```diff
@@ -86,6 +94,16 @@ export class InstanceService extends Service {
...
   }
 
+  async destroy() {
+    const keys = this.cache.keys()
+
+    for (const conduitId of keys) {
+      this.cache.del(conduitId)
+    }
+
+    this.cache.prune()
```
laurentlp, line 104: Why prune after having deleted all keys?

packages/server/src/instances/service.ts
```diff
@@ -96,18 +114,20 @@ export class InstanceService extends Service {
     try {
       await instance.initialize()
     } catch (e) {
-      instance.logger.error('Error trying to initialize conduit.', (e as Error).message)
       this.cache.del(conduitId)
 
       // TODO: replace by HealthService
```
laurentlp, line 119: What do you mean by 'replace by HealthService'. Isn't it what you indirectly do on line 126?

laurentlp (approved):
Small comments, other than that LGTM! Awesome work here :+1:

packages/server/src/instances/events.ts
samuelmasse: That's a copy pasted comment from earlier in development. I wanted to make a sort a meta event watcher but I ended up not doing that. I'll remove all these random enum values in another PR and just let the autoincremented number enums instead

packages/server/src/database/service.ts
samuelmasse: None this is pretty much a hack. But if we have queries that take over a second then we have a problem elsewhere anyways. I couldn't find a "waitForOngoingOperations()" or something function in knex so that's the best solution that works.

packages/server/src/health/service.ts
samuelmasse: Yeah I think a CallbackService or something could take care of registering important webhook calls and retrying them if they fail. I'd avoid putting that logic in WebhookService considering this service already takes care of the webhook tables. The CallbackService might have a table to keep track of failed calls to webhooks.

We won't enable this feature for messages for performance reason, but we'll need something like that for more "administrative" webhook such as health events.

packages/server/src/instances/service.ts
samuelmasse: LRU cache can have lazy behavior so here I assure that destroy() gets call on all conduits (otherwise the LRU cache might wait for the next call to `.get` before actually calling destroy.

packages/server/src/instances/service.ts
samuelmasse: Good point. I updated the comment to say "Replace by StatusService". I want to replace the tracker of failures. I initially thought this would be the health service but it'll be the StatusService instead.

PR #79, opened 2021-08-04
fix(twilio): removed unnecessary bot url for webhook validation

laurentlp:
We can use the host header instead of having to pass the BOT_URL as environment variable.

Closes MES-95

Related to: https://github.com/botpress/botpress/pull/5237


laurentlp:
We have to make sure that we merge this PR and [this](https://github.com/botpress/botpress/pull/5237) one at the same time

samuelmasse (approved):
👍

PR #81, opened 2021-08-04
fix(webhook): removed spinned url env var

laurentlp:
An instance of Botpress that starts messaging as a subprocess will also use webhooks from now on.


laurentlp:
Won't fix for now!

PR #82, opened 2021-08-04
fix(client): fix initializing a client with not auth or password

laurentlp:
Make sure to handle the case where no password or auth is provided to the client.


samuelmasse (approved):
👍

PR #89, opened 2021-08-05
fix(auth): use custom headers instead of basic auth

laurentlp:
This PR fixes an issue where multiple authentication mechanisms could conflict with our basic authentication. We will now need to provide a 'x-bp-messaging-client-*' headers to provide the clientId and clientToken.


samuelmasse (approved):
:+1:

Issue #85, opened 2021-08-05
Migration Issue: Twilio


allardy:
Twilio was working fine before the migration, but now it says that the token is invalid, but the configuration in the bot config looks fine, so maybe it's a setup issue
![image](https://user-images.githubusercontent.com/42552874/128398272-9da4554b-92df-422d-81f0-723ab12a2459.png)

samuelmasse:
@laurentlp We removed the BOT_URL env var to validate old webhooks. I think the new solution might not be working

samuelmasse:
Fixed by #94

PR #94, opened 2021-08-05
fix(twilio): fix twilio webhooks validation

laurentlp:
This PR fixes an issue where having an external URL with a port specified would not allow you to use the old webhooks with Twilio.

Related to: https://github.com/botpress/botpress/pull/5266

Closes https://github.com/botpress/messaging/issues/85


samuelmasse (approved):
:+1:

PR #95, opened 2021-08-06
feat(distributed): implement locking

samuelmasse:
Implements locking mechanism using the redlock package (recommended by redis).

The locking works a bit differently than on the main botpress repo. We try taking the lock for about 2 seconds, and if it fails it throws an error.

You can use locking like so :
```ts
// We prefix lock ids with lock_
await this.distributedService.using('lock_my_ressource', async () => {
    await doStuff()
    await doOtherStuff()
})
```

The code in the callback is mostly guaranteed to only be executed on a single node at once.

Closes MES-102

### DIFF
--- a/packages/server/src/distributed/local/subservice.ts
+++ b/packages/server/src/distributed/local/subservice.ts
@@ -24,7 +24,7 @@ export class LocalSubservice implements DistributedSubservice {
       const attempt = () => {
         const lock = this.tryGetLock(ressource)
         if (lock) {
-          return resolve(lock)
+          resolve(lock)
         } else if (attemptCount < MAX_LOCK_ATTEMPT) {
           attemptCount++
           setTimeout(attempt, LOCK_ATTEMPT_INTERVAL + Math.random() * LOCK_ATTEMPT_INTERVAL_JITTER)
--- a/packages/server/src/distributed/local/subservice.ts
+++ b/packages/server/src/distributed/local/subservice.ts
@@ -1,7 +1,7 @@
 import { DistributedSubservice } from '../base/subservice'
 import { Lock } from '../types'
 
-const DEFAULT_LOCK_TTL = 20000
+const DEFAULT_LOCK_TTL = 2000
 const MAX_LOCK_ATTEMPT = 10
 const LOCK_ATTEMPT_INTERVAL = 200
 const LOCK_ATTEMPT_INTERVAL_JITTER = 100
--- a/packages/server/src/distributed/local/subservice.ts
+++ b/packages/server/src/distributed/local/subservice.ts
@@ -2,6 +2,8 @@ import { DistributedSubservice } from '../base/subservice'
 import { Lock } from '../types'
 
 export class LocalSubservice implements DistributedSubservice {
+  private locks: { [ressource: string]: boolean } = {}
+
   async setup() {}
 
   async destroy() {}
@@ -11,8 +13,15 @@ export class LocalSubservice implements DistributedSubservice {
   async send(channel: string, message: any) {}
 
   async lock(ressource: string): Promise<Lock> {
-    return { ressource }
+    if (!this.locks[ressource]) {
+      this.locks[ressource] = true
+      return { ressource }
+    } else {
+      throw new Error('Failed to acquire lock. Already in use.')
+    }
   }
 
-  async release(lock: Lock) {}
+  async release(lock: Lock) {
+    delete this.locks[lock.ressource]
+  }
 }
--- a/packages/server/src/distributed/local/subservice.ts
+++ b/packages/server/src/distributed/local/subservice.ts
@@ -1,4 +1,5 @@
 import { DistributedSubservice } from '../base/subservice'
+import { Lock } from '../types'
 
 export class LocalSubservice implements DistributedSubservice {
   async setup() {}
@@ -8,4 +9,10 @@ export class LocalSubservice implements DistributedSubservice {
   async listen(channel: string, callback: (message: any) => Promise<void>) {}
 
   async send(channel: string, message: any) {}
+
+  async lock(ressource: string): Promise<Lock> {
+    return { ressource }
+  }
+
+  async release(lock: Lock) {}
 }
--- a/packages/server/src/distributed/local/subservice.ts
+++ b/packages/server/src/distributed/local/subservice.ts
@@ -1,8 +1,13 @@
 import { DistributedSubservice } from '../base/subservice'
 import { Lock } from '../types'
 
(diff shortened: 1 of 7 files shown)


packages/server/src/distributed/redis/subservice.ts
```diff
@@ -84,8 +88,24 @@ export class RedisSubservice implements DistributedSubservice {
...
+
+    for (const lock of Object.values(this.locks)) {
+      if (lock.expiry > now) {
+        try {
+          await this.release(lock)
+        } catch {
+          this.logger.error('Failed to release lock', _.omit(lock, 'lock'))
+        }
+      }
+    }
```
samuelmasse, line 101: Right now we hardcode every lock to have a max duration of 2 seconds. It would be possible in the future to have longer locks so this destroy() function that is executed when the server closes removes these locks. We aren't 100% precise with the timing by using new Date(), but it doesn't matter. The releasing of locks here is more for releasing locks that could still be alive for 30-40 secs.

laurentlp:
> The code in the callback is **mostly guaranteed** to only be executed on a single node at once.

**"mostly guaranteed"**? Why? Because of the timeout?

packages/server/src/distributed/local/subservice.ts
```diff
@@ -1,11 +1,27 @@
...
   async listen(channel: string, callback: (message: any) => Promise<void>) {}
 
   async send(channel: string, message: any) {}
+
+  async lock(ressource: string): Promise<Lock> {
+    if (!this.locks[ressource]) {
+      this.locks[ressource] = true
+      return { ressource }
+    } else {
+      throw new Error('Failed to acquire lock. Already in use.')
```
laurentlp, line 20: Does it mean that if you are running a single instance and try to acquire the same lock to different places, it will throw an error instead of waiting for it to be released? If so, is this what we want? 

Also, what if release is never called?

packages/server/src/distributed/redis/subservice.ts
laurentlp: Why not destroy all locks no matter if they are expired or not, when we close the server? I think it would make more sense considering that you keep an in-memory reference of all the locks!

packages/server/src/distributed/redis/subservice.ts
```diff
@@ -101,7 +121,29 @@ export class RedisSubservice implements DistributedSubservice {
     await this.pub.publish(scopedChannel, JSON.stringify({ nodeId: this.nodeId, ...message }))
   }
 
+  async lock(ressource: string) {
+    const ttl = 2000
```
laurentlp, line 125: Can you put this into a constant or read-only property!

packages/server/src/distributed/redis/subservice.ts
```diff
@@ -101,7 +121,29 @@ export class RedisSubservice implements DistributedSubservice {
...
+    const lock = {
+      lock: await this.redlock.acquire(ressource, ttl),
+      ressource,
+      expiry: new Date(new Date().getTime() + ttl)
+    }
+    this.locks[ressource] = lock
+    return lock
+  }
+
+  async release(lock: RedisLock) {
```
laurentlp, line 135: What happens if you never call release after acquiring a lock?

packages/server/src/distributed/service.ts
```diff
@@ -36,4 +37,23 @@ export class DistributedService extends Service {
...
+
+  async lock(ressource: string): Promise<Lock> {
+    return this.subservice.lock(ressource)
+  }
+
+  async release(lock: Lock) {
+    await this.subservice.release(lock)
+  }
+
+  async using(ressource: string, callback: () => Promise<void>) {
```
laurentlp, line 49: Maybe we should only expose `using` and keep `lock` and `release` private?

packages/server/src/distributed/redis/subservice.ts
samuelmasse: We don't need to release expired locks. The locks contain the expiry on redis, so other nodes checking to get a lock will check the time if the already existing lock is expired.

packages/server/src/distributed/local/subservice.ts
samuelmasse: If release is not called the lock will time out after 2 seconds. We don't need to explicitly call release. The expiry is stored on redis so other nodes can know when a lock is timed out.

We want to throw an error is some circumstances. In other we might want to skip over the bit of code if the lock exists. It depends.

packages/server/src/distributed/redis/subservice.ts
samuelmasse: The lock will time out after 2 seconds. The expiry is stored with the locks, so other nodes can know a lock is expired without having to receive a message.

samuelmasse:
I say mostly guarantees because that's what the redlock documentation says. It think the cases where it could lead to two nodes getting the same locks are some very bizarre network setup stuff.

packages/server/src/distributed/local/subservice.ts
laurentlp: > If release is not called the lock will time out after 2 seconds. We don't need to explicitly call release. The expiry is stored on redis so other nodes can know when a lock is timed out.

I understand, but what if you are not using Redis?

> We want to throw an error is some circumstances. In other we might want to skip over the bit of code if the lock exists. It depends.

Fair enough!

samuelmasse: Good point it doesn't work with the local locking. I'll put an expiry time in there instead of a boolean

packages/server/src/distributed/local/subservice.ts
```diff
@@ -1,8 +1,13 @@
 import { DistributedSubservice } from '../base/subservice'
 import { Lock } from '../types'
 
+const DEFAULT_LOCK_TTL = 20000
```
laurentlp, line 4: 20sec instead of 2 like with Redis?

samuelmasse: woops

packages/server/src/distributed/local/subservice.ts
```diff
@@ -1,11 +1,52 @@
 import { DistributedSubservice } from '../base/subservice'
+import { Lock } from '../types'
+
+const DEFAULT_LOCK_TTL = 2000
+const MAX_LOCK_ATTEMPT = 10
+const LOCK_ATTEMPT_INTERVAL = 200
```
laurentlp, line 6: Should this be: `DEFAULT_LOCK_TTL / MAX_LOCK_ATTEMPT` ?

laurentlp (approved):

PR #96, opened 2021-08-06
feat(locking): sprinkle in some locking

samuelmasse:
Adds locking to some places in the codebase.

Most particularly adds locking to instance monitoring, where not having locking could have led to bizarre scenarios. Also adds locking for sync calls, and conduit instance initializations and setups

Closes MES-105

### DIFF
--- a/packages/server/src/instances/monitoring.ts
+++ b/packages/server/src/instances/monitoring.ts
@@ -2,12 +2,14 @@ import ms from 'ms'
 import yn from 'yn'
 import { ChannelService } from '../channels/service'
 import { ConduitService } from '../conduits/service'
+import { Logger } from '../logger/types'
 import { InstanceService } from './service'
 
 const MAX_ALLOWED_FAILURES = 5
 
 export class InstanceMonitoring {
   constructor(
+    private logger: Logger,
     private channels: ChannelService,
     private conduits: ConduitService,
     private instances: InstanceService,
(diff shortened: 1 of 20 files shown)


packages/server/src/instances/monitoring.ts
```diff
@@ -10,24 +11,27 @@ const MAX_ALLOWED_FAILURES = 5
...
-      await this.loadNonLazyConduits()
+      await this.distributed.using('lock_instance_monitoring', async () => {
+        await this.initializeOutdatedConduits()
+        await this.loadNonLazyConduits()
+      })
     } catch (e) {
       this.logger.error('Error occurred while monitoring.', e.message)
+    } finally {
+      setTimeout(this.tickMonitoring.bind(this), ms('15s'))
     }
```
samuelmasse, line 35: We use a setTimeout here now instead of the more unsafe setInterval that could run mutliple times at once.

packages/server/src/instances/service.ts
```diff
@@ -153,7 +158,9 @@ export class InstanceService extends Service {
     const instance = channel.createConduit()
 
     try {
-      await instance.setup(conduitId, conduit.config, this.app)
+      await this.distributedService.using(`lock_dyn_instance_setup::${conduitId}`, async () => {
+        await instance.setup(conduitId, conduit.config, this.app)
+      })
```
samuelmasse, line 163: Something we could do here is set a lock that will last ~10sec and not release it. This would prevent different nodes from setting up the same conduit too fast in a row (which can lead on some channels to too many requests errors).

packages/server/src/instances/service.ts
```diff
@@ -117,14 +120,16 @@ export class InstanceService extends Service {
...
   }
 
   async initialize(conduitId: uuid) {
     const instance = await this.get(conduitId)
 
     try {
-      await instance.initialize()
+      await this.distributedService.using(`lock_dyn_instance_init::${conduitId}`, async () => {
+        await instance.initialize()
+      })
```
samuelmasse, line 132: Notice the `lock_dyn` prefix here. This is short for `lock_dynamic` which basically means that the name of the ressource we are locking is dynamic (as opposed to a static lock name such as `lock_instance_monitoring` that we use in the monitoring). We should keep this a standard in the code to make searching easier (basically the same principle as `cache_` and `batcher_`)

packages/server/src/instances/monitoring.ts
```diff
@@ -10,24 +11,27 @@ const MAX_ALLOWED_FAILURES = 5
...
-      await this.initializeOutdatedConduits()
-      await this.loadNonLazyConduits()
+      await this.distributed.using('lock_instance_monitoring', async () => {
+        await this.initializeOutdatedConduits()
+        await this.loadNonLazyConduits()
+      })
     } catch (e) {
       this.logger.error('Error occurred while monitoring.', e.message)
+    } finally {
+      setTimeout(this.tickMonitoring.bind(this), ms('15s'))
```
laurentlp, line 34: Aren't we exposed to problems if `initializeOutdatedConduits` and `loadNonLazyConduits` take more than 15sec to complete?

samuelmasse: No the locking will throw an exception and the node will just skip monitoring for that time. We just want to avoid multiple nodes doing the monitoring at the same time

laurentlp (approved):
LGTM! :+1:

PR #97, opened 2021-08-06
chore(app): failsafe destruction of services

samuelmasse:
Since we want to free up as much stuff as we can when the server closes/crashes, we also want to avoid this freeing up of stuff to also crash and prevent further stuff from being freed.

That's why I refactored the `destroy` functions of services a bit to make them as much failsafe as possible so that all services get to destroy their stuff (if they can).

Closes MES-109

### DIFF
--- a/packages/server/src/instances/service.ts
+++ b/packages/server/src/instances/service.ts
@@ -110,12 +110,13 @@ export class InstanceService extends Service {
   }
 
   async destroy() {
-    const keys = this.cache.keys()
-
-    for (const conduitId of keys) {
-      this.cache.del(conduitId)
+    if (!this.cache) {
+      return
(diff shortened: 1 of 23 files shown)


packages/server/src/instances/service.ts
```diff
@@ -110,12 +110,13 @@ export class InstanceService extends Service {
   }
 
   async destroy() {
-    const keys = this.cache.keys()
+    if (!this.cache) {
+      return
+    }
```
samuelmasse, line 115: Note that it's possible for the server to crash in the constructor so cache can be undefined in that scenario

laurentlp (approved):
LGTM! Nice work!!!!

PR #98, opened 2021-08-06
chore(logging): refactor error logging

samuelmasse:
The error logging was a bit all over the place. So now there is a standard way of passing the Error object to the logger.

Closes MES-110

### DIFF
--- a/packages/server/src/channels/messenger/channel.ts
+++ b/packages/server/src/channels/messenger/channel.ts
@@ -89,7 +89,9 @@ export class MessengerChannel extends Channel<MessengerConduit> {
     const [, hash] = signature.split('=')
     const expectedHash = crypto.createHmac('sha1', conduit.config.appSecret).update(buffer).digest('hex')
     if (hash !== expectedHash) {
-      this.logger.error("Couldn't validate the request signature. Make sure your appSecret is properly configured.")
+      // TODO: can we just throw the error with the message in it instead??
+      const e = new Error()
+      this.logger.error(e, "Couldn't validate the request signature. Make sure your appSecret is properly configured")
 
       throw new Error()
     }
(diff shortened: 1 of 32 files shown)


packages/server/src/channels/messenger/channel.ts
```diff
@@ -89,7 +89,9 @@ export class MessengerChannel extends Channel<MessengerConduit> {
     const [, hash] = signature.split('=')
     const expectedHash = crypto.createHmac('sha1', conduit.config.appSecret).update(buffer).digest('hex')
     if (hash !== expectedHash) {
-      this.logger.error("Couldn't validate the request signature. Make sure your appSecret is properly configured.")
+      // TODO: can we just throw the error with the message in it instead??
+      const e = new Error()
+      this.logger.error(e, "Couldn't validate the request signature. Make sure your appSecret is properly configured")
```
laurentlp, line 94: Yeah, I think you should do something like this:

```suggestion
      const message = "Couldn't validate the request signature. Make sure your appSecret is properly configured"
      this.logger.error(new Error(message), message)
```

So at least you have the stack trace and the error message if needed!

packages/server/src/channels/twilio/channel.ts
```diff
@@ -37,7 +37,8 @@ export class TwilioChannel extends Channel<TwilioConduit> {
           await this.app.instances.receive(conduit.conduitId, req.body)
           res.sendStatus(204)
         } else {
-          this.logger.error('Request validation failed. Make sure that your authToken is valid.')
+          // Maybe throw a exception here. Or just do new Error() and log that?
+          this.logger.error(undefined, 'Request validation failed. Make sure that your authToken is valid')
```
laurentlp, line 41: I think creating a new error is fine. It will require less code than throwing an exception and is still easy to read!

packages/server/src/channels/messenger/channel.ts
samuelmasse: I'll just log the error, so we avoid having the same message twice

laurentlp (approved):

PR #99, opened 2021-08-06
feat(telegram): restore polling

samuelmasse:
This PR restores the polling feature of telegram so that it can be used in single node environment when spinned from botpress.

Closes MES-99


laurentlp (approved):
LGTM!

PR #100, opened 2021-08-06
fix(smooch): fix webhook setup

samuelmasse:
Fixes issue where smooch would produce an error for recreating a webhook that already existed. We also hardcode routes that contain `/mod/channel-smooch` to be deleted to avoid messages being doubled sent

Closes MES-112

### DIFF
--- a/packages/server/src/channels/smooch/conduit.ts
+++ b/packages/server/src/channels/smooch/conduit.ts
@@ -1,5 +1,6 @@
 // @ts-ignore
 import Smooch from 'smooch-core'
+import yn from 'yn'
 import { ConduitInstance, EndpointContent } from '../base/conduit'
 import { ChannelContext } from '../base/context'
 import { CardToCarouselRenderer } from '../base/renderers/card'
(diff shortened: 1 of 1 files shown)


packages/server/src/channels/smooch/conduit.ts
```diff
@@ -24,24 +46,12 @@ export class SmoochConduit extends ConduitInstance<SmoochConfig, SmoochContext>
...
     const target = await this.getRoute()
+    const webhook = webhooks.find((x: any) => x.target === target)
+    this.secret = webhook?.secret
 
-    try {
-      // Note: creating a webhook with the same url will not create a new webhook but return the already existing one
-      const { webhook }: { webhook: SmoochWebhook } = await this.smooch.webhooks.create({
-        target,
-        triggers: ['message:appUser']
-      })
```
samuelmasse, line 39: I removed the try catch here because it is catched by the try catch before instance.initialize() in InstanceService. It's actually necessary to throw an exception here otherwise we don't know that the setup failed

packages/server/src/channels/smooch/conduit.ts
```diff
@@ -14,7 +15,28 @@ export class SmoochConduit extends ConduitInstance<SmoochConfig, SmoochContext>
   public secret!: string
 
   async initialize() {
-    await this.setupWebhook()
+    const target = await this.getRoute()
+    const { webhooks } = await this.smooch.webhooks.list()
+
+    if (yn(process.env.SPINNED)) {
```
laurentlp, line 21: Why not delete it even if we use an external messaging server?

samuelmasse: Because that's a useless operation that we don't want to be doing on the messaging server. We only do this so migrations go well. Putting your bot on the cloud is more than a migration.

laurentlp (approved):

PR #105, opened 2021-08-09
feat(webhook): common sender class

laurentlp:
This PR put all the duplicated to that was used to call webhooks into one class.

### DIFF
--- a/packages/server/src/health/service.ts
+++ b/packages/server/src/health/service.ts
@@ -13,6 +13,7 @@ import { DatabaseService } from '../database/service'
 import { InstanceService } from '../instances/service'
 import { LoggerService } from '../logger/service'
 import { Logger } from '../logger/types'
+import { WebhookBroadcaster } from '../webhooks/broadcaster'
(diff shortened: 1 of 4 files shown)


samuelmasse (approved):
Ok good job that's much cleaner :+1: . This will enventually be replace by a `PostService` (which will have all the retries and such). It could also be named `BroadcasterService`. For now this is simple enough to keep in an utility class.

packages/server/src/health/service.ts
```diff
@@ -39,6 +41,7 @@ export class HealthService extends Service {
     this.table = new HealthTable()
     this.watcher = new HealthWatcher(this.conduitService, this.instanceService, this)
     this.logger = this.loggerService.root.sub('health')
+    this.webhookBroadcaster = new WebhookBroadcaster(this.configService, this.webhookService)
```
samuelmasse, line 44: `webhookBroadcaster` is kind on unnecessarily long. You don't have really anything else here that could be thought of as a broadcaster. I would just name this variable `broadcaster`

Issue #101, opened 2021-08-06
Suggestion: Do not return channel id in webhook post back


EFF:
**Currently**: When we capture a user's message in the webhook we receive something like the following
```
{
 channelId: 'asdasdas',
 clientId: 'asdasd',
 /*....*/
}
```

Then if you want to reply to that message you need to specify the channel and no where it says that we should use the channel name and not the id which can become confusing.  It's also super tempting to use `data.channel`.

From what I understand, channel-id is for internal usage so it shouldn't be exposed to the user

samuelmasse:
Agreed. Channel id is for internal usage so sending it to the api user is pretty confusing. It would also be better not to have to supply the channel name at all when making a reply (which should be possible to implement pretty easily).

PR #106, opened 2021-08-09
feat(webhook): do not send channel id

laurentlp:
Closes https://github.com/botpress/messaging/issues/101

I also added some debug logging like so: 
Screenshot from 2021-08-09 16-44-02

### DIFF
--- a/packages/server/src/webhooks/broadcaster.ts
+++ b/packages/server/src/webhooks/broadcaster.ts
@@ -2,24 +2,30 @@ import axios, { AxiosRequestConfig } from 'axios'
 import yn from 'yn'
 
 import { ConfigService } from '../config/service'
+import { Logger } from '../logger/types'
 import { WebhookService } from './service'
+import { WebhookContent } from './types'
 
 export class WebhookBroadcaster {
-  constructor(private configService: ConfigService, private webhookService: WebhookService) {}
+  private logger: Logger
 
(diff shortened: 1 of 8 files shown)


packages/server/src/webhooks/broadcaster.ts
```diff
@@ -32,6 +38,8 @@ export class WebhookBroadcaster {
     }
 
     try {
+      this.logger.debug(`Client ID '${clientId}' calling webhook URL '${url}'`)
```
samuelmasse, line 41: You need to check if logging is enabled.

```ts
if (yn(process.env.LOGGING_ENABLED)) {
  this.logger.debug(`Client ID '${clientId}' calling webhook URL '${url}'`)
}
```

samuelmasse (commented):
Why not send client ids? You could have the same route configured to listen to 2 different client's events.

packages/server/src/webhooks/broadcaster.ts
laurentlp: It would be better if this logic was inside the logger itself so we don't have to make that check each time we call debug!

samuelmasse: We'll do something similar to that later. For now it's necessary to check if logging is enabled that way otherwise botpress will be bombarded with logs

samuelmasse (approved):
:+1:

PR #109, opened 2021-08-09
fix(server): run messaging in production when using the binary

laurentlp:
This PR sets the value of NODE_ENV to production when running a binary version of the messaging server.

Closes MES-24

### DIFF
--- a/packages/server/src/index.ts
+++ b/packages/server/src/index.ts
@@ -3,6 +3,9 @@ import { Api } from './api'
 import { App } from './app'
 import { Launcher } from './launcher'
 
+// Set NODE_ENV to production when starting messaging using the binary version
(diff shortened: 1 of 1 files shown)


packages/server/src/index.ts
```diff
@@ -3,6 +3,9 @@ import { Api } from './api'
 import { App } from './app'
 import { Launcher } from './launcher'
 
+// Set NODE_ENV to production when starting messaging using the binary version
+process.env.NODE_ENV = (<any>process).pkg ? 'production' : process.env.NODE_ENV
```
laurentlp, line 7: process.pkg is only defined when the server has been bundled by the pkg library!

samuelmasse (approved):
:+1:

PR #117, opened 2021-08-10
chore(post): added unit tests for post service

laurentlp:
This PR adds unit tests for the post service and adds a launch config for debugging unit tests.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -7,5 +7,6 @@
   "eslint.nodePath": "./node_modules",
   "editor.codeActionsOnSave": {
     "source.fixAll": true
-  }
+  },
+  "jest.jestCommandLine": "yarn test"
 }
--- a/packages/server/tests/post/service.test.ts
+++ b/packages/server/tests/post/service.test.ts
@@ -1,0 +1,136 @@
+import _ from 'lodash'
+import axios, { AxiosRequestConfig } from 'axios'
+
+import { PostService } from '../../src/post/service'
+import { ConfigService } from '../../src/config/service'
+
+jest.mock('../../src/config/service')
+jest.mock('../../src/logger/types')
+jest.mock('axios')
+
+describe('PostService', () => {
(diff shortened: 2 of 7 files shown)


packages/server/tests/post/service.test.ts
```diff
@@ -0,0 +1,136 @@
...
+    configService['current'] = {}
+
+    data = { some: 'data' }
+    axiosConfig = { headers: {} }
+    headers = { custom: 'header' }
+  })
+
+  afterEach(() => {
+    jest.clearAllMocks()
+  })
```
samuelmasse, line 31: You can remove that with the new config

packages/server/tests/post/service.test.ts
```diff
@@ -0,0 +1,136 @@
...
+
+  it('Should instantiate without throwing any error', async () => {
```
samuelmasse, line 33: Please rename all it to test

.vscode/settings.json
```diff
@@ -7,5 +7,6 @@
   "eslint.nodePath": "./node_modules",
   "editor.codeActionsOnSave": {
     "source.fixAll": true
-  }
+  },
+  "jest.jestCommandLine": "yarn test"
```
samuelmasse, line 11: What does this do?

samuelmasse (approved):
:+1: just make sure to adjust to the new test setup

EFF:
I'd love if we could start using conventional commits properly. unit tests are not features .. clearly it's a chore

A feature is something that is exposed to the user, a new functionality

.vscode/settings.json
laurentlp: It makes sure that the Jest vscode extension uses yarn test instead of using the jest binary.

packages/server/tests/post/service.test.ts
laurentlp: Will be done in the base branch

PR #118, opened 2021-08-10
feat(post): added exponential backoff support

laurentlp:
This PR adds exponentials backoff as the retry strategy for the post service. The default number of attempts is 10, but this could be easily configured. I'm open to ideas/criticism!

The formula for the amount of time it gonna take if the call never resolves is: `100 * 2^(n-1)` (where `n` is the number of retries) 

e.g.

With a [jitter](https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/):

Screenshot from 2021-08-10 16-16-27

Without a jitter:

Screenshot from 2021-08-10 16-18-05

Closes MES-122

### DIFF
--- a/packages/server/src/post/service.ts
+++ b/packages/server/src/post/service.ts
@@ -1,10 +1,13 @@
 import axios, { AxiosRequestConfig } from 'axios'
-import { Service } from '../base/service'
+import { backOff } from 'exponential-backoff'
 
+import { Service } from '../base/service'
 import { ConfigService } from '../config/service'
 import { Logger } from '../logger/types'
 
 export class PostService extends Service {
+  private readonly attempts = 10
+
   private logger: Logger
   private password: string | undefined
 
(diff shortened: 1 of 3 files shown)


packages/server/src/post/service.ts
```diff
@@ -29,10 +32,18 @@ export class PostService extends Service {
     }
 
     try {
-      await axios.post(url, data, config)
+      await backOff(async () => axios.post(url, data, config), {
+        jitter: 'none', // TODO: We should enable to jitter if the post service may be called for the same resource on multiple nodes
```
samuelmasse, line 36: What it this comment about? I don't really understand what you're saying.

samuelmasse (approved):
:+1:

packages/server/src/post/service.ts
laurentlp: If the post service, for some reason may call the same URL at the same time and on multiple instances of the messaging server, we should enable the jitter!

PR #119, opened 2021-08-10
chore(tests): improve test setup

samuelmasse:
This PR brings some changes to the setup of jest.

1. I moved the tests from the `tests` folder to the `test` folder. This is very nitpicky but I just felt like the "s" wasn't adding anything to the folder name.
2. I changed the jest.config file to use typescript. This gives use typings. I also deleted the tsconfig.test.json file.
3. I use the `test` function istead of `it` everywhere for tests. `it` is kind of a cryptic name for what it does.
4. I added a `clearMocks: true` to the jest.config file that allows use to delete the `jest.clearAllMocks()` code from each test by doing it automatically.

We can merge this now. Additionally I'll look into having our mock objects in .mock.ts files instead of the magic __mocks __ folder.

There is also a VERY annoying problem at the moment with vscode. It seems vscode refuses to suggest imports of code that is in the src folder when writing test. This makes test writing way slower.


laurentlp (approved):
Nice work :+1:

PR #120, opened 2021-08-10
feat(client): add support for custom http clients

laurentlp:
This PR adds support to instantiate the MessagingClient with a custom axios instance giving more control to the user over the internal HTTP client.

Closes MES-124


samuelmasse:
Why not take an AxiosConfig as an optional parameter in MessagingOptions instead? The field could simply be called `axios` or `axiosConfig`

samuelmasse:
Also we can remove the password option now. We'll just pass that option by setting the headers option instead.

samuelmasse (approved):
:+1:

PR #125, opened 2021-08-12
chore(database): improve unit tests for the database service

laurentlp:
This PR:

 1. Improves the unit tests for the database service
 2. Makes sure that everything is properly destroyed when closing the server


packages/server/src/instances/service.ts
```diff
@@ -97,7 +99,7 @@ export class InstanceService extends Service {
     this.lazyLoadingEnabled = !yn(process.env.NO_LAZY_LOADING)
 
     this.cache = await this.cachingService.newServerCache('cache_instance_by_conduit_id', {
-      dispose: this.handleCacheDispose.bind(this),
+      dispose: () => !this.destroyed && this.handleCacheDispose.bind(this),
```
samuelmasse, line 102: I don't think this works. You're not calling the function here.
```ts
dispose: async (k, v) => {
        if (!this.destroyed) {
          await this.handleCacheDispose(k, v)
        }
      },
```

samuelmasse (approved):
There's a bug in the cache dispose function otherwise nice work :+1:

samuelmasse:
Ok I tested it and it and it works. I added a small fix commit. Basically we don't need to clear the cache anymore.

It seems like the http server takes a few seconds before resolving the promise. I don't know if that's a timeout or something else but we'll prefer that to our own random timer.

PR #128, opened 2021-08-13
fix(telegram): destroy telegraf

samuelmasse:
This PR adds a stop call to telegraf when the instance is destroyed. This prevents a bug where a deleted conduit would have an instance that keeps polling messages and tries to receive messages with that conduit.


laurentlp (approved):
Good catch! :+1:

PR #129, opened 2021-08-13
chore(base): implement base package

samuelmasse:
Add a package name `base` that contains typings used by both the client and the server.


package.json
```diff
@@ -13,9 +13,9 @@
     "release": "node ./scripts/release",
     "build": "yarn --silent && tsc --build",
     "watch": "yarn --silent && tsc --build --watch",
-    "start": "cd packages/server && node dist/index.js",
-    "dev": "cd packages/server && yarn --silent && ts-node-dev --inspect --inspect-publish-uid=http --transpile-only src/index.ts",
-    "board": "cd packages/board && yarn dev",
+    "start": "yarn workspace @botpress/messaging-server start",
+    "dev": "yarn workspace @botpress/messaging-server dev",
+    "board": "yarn workspace @botpress/messaging-board dev",
```
laurentlp, line 18: We could maybe have these commands instead:

```suggestion
    "server:start": "yarn workspace @botpress/messaging-server start",
    "server:dev": "yarn workspace @botpress/messaging-server dev",
    "board:dev": "yarn workspace @botpress/messaging-board dev",
    ...
```

What do you think?

packages/base/package.json
```diff
@@ -0,0 +1,17 @@
...
+  "license": "AGPL-3.0",
+  "scripts": {
+    "build": "yarn --silent && tsc --build",
+    "watch": "yarn --silent && tsc --build --watch"
+  },
+  "files": [
+    "dist"
+  ],
+  "devDependencies": {},
+  "dependencies": {}
```
laurentlp, line 16: Shouldn't this package depend on typescript?

packages/client/package.json
```diff
@@ -15,6 +15,7 @@
   ],
   "devDependencies": {},
   "dependencies": {
-    "axios": "^0.21.1"
+    "axios": "^0.21.1",
+    "@botpress/messaging-base": "0.0.1"
```
laurentlp, line 19: Since this package only contains typings I think it should be in the devDependencies section!

packages/client/src/chat.ts
```diff
@@ -1,8 +1,15 @@
...
 import { BaseClient } from './base'
-import { Message } from './messages'
 
 export class ChatClient extends BaseClient {
   async reply(conversationId: string, channel: string, payload: any): Promise<Message> {
-    return (await this.http.post('/chat/reply', { conversationId, channel, payload })).data
+    return this.deserialize((await this.http.post('/chat/reply', { conversationId, channel, payload })).data)
+  }
+
+  private deserialize(message: any) {
```
laurentlp, line 9: Do you have an idea of what we could use here instead of `any`? `object`?

I'm asking because currently, nothing prevents you from calling deserialize with a string which would result in an error I think!

packages/client/package.json
```diff
@@ -15,6 +15,7 @@
   ],
   "devDependencies": {},
   "dependencies": {
```
laurentlp, line 17: Can you bump the package version and publish a new version of the client?!

laurentlp (approved):
Small comments other than that LGTM! Good job, I really love this refactoring!! :muscle:

packages/base/package.json
samuelmasse: Typescript is a dev dependency of the whole workspace so no

package.json
samuelmasse: No I like the short commands. If we ever get to to many commands maybe, but 3 isn't hard to remember

packages/client/package.json
samuelmasse: The package contains a emitter class that implements functions so it's not just a types package

packages/client/src/chat.ts
samuelmasse: I think any is fine. It's a private function anyway so it's not that dangerous to misuse

packages/client/package.json
samuelmasse: I'll do that later when we deploy the webchat

packages/client/src/chat.ts
laurentlp: There are places where we expose a similar deserialize function where the parameter is also of type any tho!

PR #131, opened 2021-08-15
fix(server): added server manager to handle closing server connections

laurentlp:
This PR adds a check to make sure we manually destroy all the Express server connections when the Messaging server is shutting down.

Should fix the issue where debugging in development mode would result in the debugger port already being taken.

This code is based on libraries recommended by Express to handle graceful shutdowns: https://expressjs.com/en/advanced/healthcheck-graceful-shutdown.html


samuelmasse (approved):
Works great nice work!

PR #133, opened 2021-08-15
feat(webhook): make sure webhook urls are valid

laurentlp:
This PR adds a check to make sure that the synced webhook URLs are valid URIs and are reachable using an option request.

Closes MES-125


samuelmasse:
It looks like there is a problem with Jest and importing other packages from yarn workspaces. 

```
src/channels/base/channel.ts:1:22 - error TS2307: Cannot find module '@botpress/messaging-base' or its corresponding type declarations.

    1 import { uuid } from '@botpress/messaging-base'
                           ~~~~~~~~~~~~~~~~~~~~~~~~~~
```

samuelmasse:
Building the repo first seems to solve the problem. I think rewiring might also solve it. You can see how the rewiring works when working with ts-node in the rewire.ts. Ideally we would need to require that file before every test.

packages/server/src/sync/service.ts
```diff
@@ -198,9 +199,16 @@ export class SyncService extends Service {
...
+        try {
+          await axios.options(webhook.url)
+
+          const token = await this.webhooks.generateToken()
+          const newToken = await this.webhooks.create(clientId, token, webhook.url)
+          webhooksWithTokens.push({ url: newToken.url, token: newToken.token })
+        } catch {
+          // TODO: Send this info back to the user
+          this.logger.warn(`Webhook with url "${webhook.url}" cannot be reached. Skipping...`)
+        }
```
samuelmasse, line 211: Yup we need to return a 400 error with that messages. The sync shouldn't happen if the webhook is unreachable otherwise it kind of confusing to have the webhooks be scretly not set.

samuelmasse (changes requested):
We need to send a 400 with a message back to the user otherwise great work!

packages/server/src/sync/service.ts
samuelmasse: So to avoid other changes being done, we'd need to have a verify webhooks function before everything else.

laurentlp: Since we seem to sync bots before the /receive API is up, I had to remove this validation! We will simply make sure that the URL sent is of valid format using Joi.

laurentlp:
> Building the repo first seems to solve the problem. I think rewiring might also solve it. You can see how the rewiring works when working with ts-node in the rewire.ts. Ideally we would need to require that file before every test.

https://github.com/kulshekhar/ts-jest/issues/1648

samuelmasse (approved):
:+1:

PR #134, opened 2021-08-16
feat(dx): add .nvmrc

spg:
`.nvmrc` is useful when using `nvm`. When a `.nvmrc` file is detected in the current working directory, `nvm` will use the Node version specified in the `.nvmrc` file.


laurentlp (commented):
Not sure if the code is compatible with NodeJS 16. We would have to test the build, tests, and binaries first!

spg:
What's the current version of Node you're supporting?

samuelmasse:
@spg If you look at the git changes we already had a .nvmrc with 12.18.3 in it

spg:
Oh yeah sorry, my bad

PR #136, opened 2021-08-17
fix(logger): display stack traces on one line

laurentlp:
This PR makes sure that we display stack traces on one line so that any log aggregator can properly differentiate between stack traces.

Maybe we should use a `CLOUD` (or some other) env var to tell messaging that we prefer to output single-line stack traces and keep the multi-line traces otherwise? What do you think?

Closes MES-131


spg:
Personally I prefer more fine-grained environment variables, e.g. `MULTILINE_LOGGING=false`. Using a `CLOUD` environment variable can mean a lot of things.

samuelmasse (approved):
:+1:

samuelmasse:
This PR is marked as a draft but I assume it's ready. This happens often can you make sure to put your PR as ready for review when they are

laurentlp:
> This PR is marked as a draft but I assume it's ready. This happens often can you make sure to put your PR as ready for review when they are

I'm still working on it, that's why it's a draft!

PR #137, opened 2021-08-17
fix(dockerfile): adapt dockerfile to project references

laurentlp:
This PR fixes an issue where the @botpress/messaging-base package could not be found due to the way the dockerfile previously built the project.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,3 +18,34 @@ jobs:
       - name: Build Messaging
         run: |
           yarn build
+
+  # Taken from: https://github.com/docker/build-push-action/blob/master/docs/advanced/cache.md#local-cache
+  docker:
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -18,3 +18,34 @@ jobs:
...
         run: |
           yarn build
+
+  # Taken from: https://github.com/docker/build-push-action/blob/master/docs/advanced/cache.md#local-cache
+  docker:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Set up Docker Buildx
```
spg, line 28:
```suggestion
      - name: Set up Docker Build
```

spg (approved):

PR #140, opened 2021-08-17
fix(post): stop retrying request when server is shutting down

laurentlp:
This PR fixes an issue where the post service could still retry multiple times a certain request when the server is shutting down.


samuelmasse (approved):
:+1:

PR #141, opened 2021-08-17
fix(server): added http-terminator as server manager and removed custom one

laurentlp:
This PR adds http-terminator as server manager to handle graceful shutdowns with ExpressJS.


samuelmasse:
Doesn't build `src/launcher.ts(11,31): error TS2307: Cannot find module './server-manager' or its corresponding type declarations.`

laurentlp:
> Doesn't build `src/launcher.ts(11,31): error TS2307: Cannot find module './server-manager' or its corresponding type declarations.`

Fixed

samuelmasse (approved):
:+1:

PR #143, opened 2021-08-20
feat(server): gracefully close websocket connections on server shutdown

laurentlp:
This PR makes sure to properly close the websocket server and its connections upon server shutdown.

Closes MES-132

### DIFF
--- a/packages/server/src/launcher.ts
+++ b/packages/server/src/launcher.ts
@@ -83,11 +83,17 @@ export class Launcher {
     if (!this.shuttingDown) {
       this.shuttingDown = true
 
-      this.logger.info('Server gracefully closing down...')
+      if (!yn(process.env.SPINNED)) {
+        this.logger.info('Server gracefully closing down...')
+      }
 
       await this.api.sockets.destroy()
       await this.httpTerminator?.terminate()
       await this.app.destroy()
+
+      if (!yn(process.env.SPINNED)) {
+        this.logger.info('Server shutdown complete')
+      }
     }
     process.exit(code)
   }
--- a/packages/server/src/launcher.ts
+++ b/packages/server/src/launcher.ts
@@ -83,6 +83,9 @@ export class Launcher {
     if (!this.shuttingDown) {
       this.shuttingDown = true
 
+      this.logger.info('Server gracefully closing down...')
+
+      await this.api.sockets.destroy()
       await this.httpTerminator?.terminate()
       await this.app.destroy()
     }
--- a/packages/server/src/socket/manager.ts
+++ b/packages/server/src/socket/manager.ts
@@ -7,15 +7,36 @@ import { SocketService } from './service'
 
 export class SocketManager {
   private logger = new Logger('Socket')
+  private ws: Socket.Server | undefined
   private handlers: { [type: string]: SocketHandler } = {}
 
   constructor(private sockets: SocketService) {}
 
   async setup(server: Server) {
     if (yn(process.env.ENABLE_EXPERIMENTAL_SOCKETS)) {
-      const ws = new Socket.Server(server, { cors: { origin: '*' } })
-      ws.on('connection', this.handleSocketConnection.bind(this))
+      this.ws = new Socket.Server(server, { cors: { origin: '*' } })
+      this.ws.on('connection', this.handleSocketConnection.bind(this))
     }
+  }
+
+  async destroy() {
+    if (this.ws !== undefined) {
+      await new Promise((resolve, reject) => {
+        // This is kind of hack to make sure that socket.io does not
+        // try to close the HTTP server before http-terminator
+        this.ws!['httpServer'] = undefined
+
+        this.ws!.close((err) => {
+          if (err) {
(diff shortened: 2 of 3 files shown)


packages/server/src/launcher.ts
```diff
@@ -83,6 +83,9 @@ export class Launcher {
     if (!this.shuttingDown) {
       this.shuttingDown = true
 
+      this.logger.info('Server gracefully closing down...')
+
```
samuelmasse, line 87: Only print this if we're not spinned from botpress

`if (!yn(process.env.SPINNED)) {`

Also I think a message `Shutdown complete` or something after would be useful.

packages/server/src/socket/manager.ts
```diff
@@ -7,17 +7,38 @@ import { SocketService } from './service'
...
+      this.ws.on('connection', this.handleSocketConnection.bind(this))
     }
   }
 
+  async destroy() {
+    if (this.ws !== undefined) {
+      await new Promise((resolve, reject) => {
+        // This is kind of hack to make sure that socket.io does not
+        // try to close the HTTP server before http-terminator
+        this.ws!['httpServer'] = undefined
```
samuelmasse, line 27: Did this happen while testing?

packages/server/src/socket/manager.ts
```diff
@@ -7,17 +7,38 @@ import { SocketService } from './service'
...
+          if (err) {
+            return reject(err)
+          }
+
+          resolve(undefined)
+        })
+      })
+    }
+
+    await this.sockets.destroy()
```
samuelmasse: This should be in app.destroy() where we destroy all the other services.

packages/server/src/socket/service.ts
```diff
@@ -21,6 +21,19 @@ export class SocketService extends Service {
...
+      if (!sockets) {
+        continue
+      }
+
+      for (const socket of sockets) {
+        // TODO: Delete the socket from the cache and invalidate once we distribute the socket usage
+        socket.disconnect()
+      }
+    }
+  }
```
samuelmasse: You can use `this.sockets` instead to list all sockets.

packages/server/src/socket/manager.ts
```diff
@@ -7,17 +7,38 @@ import { SocketService } from './service'
...
     if (yn(process.env.ENABLE_EXPERIMENTAL_SOCKETS)) {
-      const ws = new Socket.Server(server, { cors: { origin: '*' } })
-      ws.on('connection', this.handleSocketConnection.bind(this))
+      this.ws = new Socket.Server(server, { cors: { origin: '*' } })
+      this.ws.on('connection', this.handleSocketConnection.bind(this))
     }
   }
 
+  async destroy() {
+    if (this.ws !== undefined) {
```
samuelmasse, line 23:
```ts
if (!this.ws) {
  return
}
```

packages/server/src/socket/manager.ts
```diff
@@ -7,17 +7,38 @@ import { SocketService } from './service'
...
+      await new Promise((resolve, reject) => {
+        // This is kind of hack to make sure that socket.io does not
+        // try to close the HTTP server before http-terminator
+        this.ws!['httpServer'] = undefined
+
+        this.ws!.close((err) => {
+          if (err) {
+            return reject(err)
+          }
```
samuelmasse, line 32: It would be a good idea to log the error here but don't reject the promise.  Destroy functions need to be as failsafe as possible.

samuelmasse (commented):
I'm wondering if any of this is necessary. If we remove the http server from the Socket Server before calling close, then what does the close function do?

packages/server/src/socket/manager.ts
laurentlp: This call was removed as it is unnecessary!

packages/server/src/socket/manager.ts
laurentlp: Yes! Calling close on the server when it is stopped throws an unexpected exception. Since socket.io calls close on the server if it is defined (https://github.com/socketio/socket.io/blob/master/lib/index.ts#L652) and then http-terminator does the same, an exception is thrown.

packages/server/src/socket/service.ts
laurentlp: This function was removed as it is unnecessary!

packages/server/src/socket/manager.ts
samuelmasse: Ok cool looking at the code this should work perfeclty

laurentlp:
> I'm wondering if any of this is necessary. If we remove the http server from the Socket Server before calling close, then what does the close function do?

I agree that it is kind of unnecessary, but still, it seems to be cleaner that way and doesn't add overhead. If you take a look at the definition of the server's [`close` function](https://github.com/socketio/socket.io/blob/master/lib/index.ts#L644), you can see that it does more than just closing the HTTP server.

samuelmasse (approved):
:+1:

PR #145, opened 2021-08-20
feat(board): host url field

samuelmasse:
Adds a field to the board the allows changing the host of the webchat in the ui instead of having to do it in the code.
Screenshot from 2021-08-20 17-36-57

### DIFF
--- a/packages/board/src/linker.ts
+++ b/packages/board/src/linker.ts
@@ -88,10 +88,7 @@ export class BoardLinker {
 
     let host = this.inputHost.value
     if (!host?.length) {
-      host = localStorage.getItem('bp-host')!
-    }
-    if (!host?.length) {
-      host = 'http://localhost:3100'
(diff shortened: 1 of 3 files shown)


packages/board/src/linker.ts
```diff
@@ -79,12 +86,20 @@ export class BoardLinker {
   private async create() {
     this.webchat?.destroy()
 
+    let host = this.inputHost.value
+    if (!host?.length) {
+      host = localStorage.getItem('bp-host')!
+    }
+    if (!host?.length) {
+      host = 'http://localhost:3100'
+    }
```
laurentlp, line 95: Why not simply do:

```suggestion
    const DEFAULT_HOST = 'http://localhost:3100'
    // ...
    if (!host) {
      host = localStorage.getItem('bp-host') || DEFAULT_HOST
    }
```

laurentlp (approved):
Small comment, else LGTM!!! :muscle:

PR #146, opened 2021-08-23
fix(ci_cd): fix changelog generation on release

laurentlp:
This PR fixes CHANGELOG generation on new release. 

The process is now as follows:

```sh
// Create a new branch from an up-to-date master
git checkout -b vX_X_X
// Run yarn release to bump the package.json version and generate a changelog
yarn release <major|minor|patch>
// Commit the changes and push the branch and create a pull request
git add .
git commit -m "vX_X_X"
git push origin vX_X_X
```

See https://github.com/botpress/messaging/pull/147

Closes MES-37

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -14,7 +14,7 @@ jobs:
         with:
           node-version: '12.18.3'
           cache: 'yarn'
-      - name: Git Unshalow
+      - name: Git Unshallow
         run: git fetch --prune --unshallow
       - name: Fetch Node Packages
         run: |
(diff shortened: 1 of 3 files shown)


samuelmasse (approved):
Good job :+1: 

We should look into rewriting all the code in the scripts folder. I don't see any reason why it has to be in javascript. We could use ts-node instead.

laurentlp:
> Good job 
> 
> We should look into rewriting all the code in the scripts folder. I don't see any reason why it has to be in javascript. We could use ts-node instead.

Totally agree, I created an issue to work on this later: https://linear.app/botpress/issue/MES-136/convert-scripts-to-ts-files

PR #152, opened 2021-08-24
chore(config): remove json config file

samuelmasse:
Doc update is missing

Removes the config.json file from the code. Everything is now loaded from environment variables.

Closes MES-137


packages/server/src/database/service.ts
```diff
@@ -3,23 +3,17 @@ import knex, { Knex } from 'knex'
 import path from 'path'
 import { Service } from '../base/service'
 import { Table } from '../base/table'
-import { ConfigService } from '../config/service'
 import { Logger } from '../logger/types'
 
 export class DatabaseService extends Service {
   public knex!: Knex
   private url!: string
```
laurentlp, line 10:
```suggestion
  private url!: string | undefined
```

Then you can remove the exclamation point at the end of `process.env.DATABASE_URL` on line 16.

packages/server/src/database/service.ts
```diff
@@ -30,7 +24,7 @@ export class DatabaseService extends Service {
   }
 
   private loadPoolConfig() {
-    let config = this.configService.current.database?.pool
+    let config = undefined
```
laurentlp, line 27: IMHO, you could remove this line and simply put line 37 inside the try-catch. It would be cleaner that way

packages/server/src/sync/service.ts
```diff
@@ -33,7 +31,7 @@ export class SyncService extends Service {
   }
 
   async setup() {
-    let config = this.config.current.sync
+    let config = undefined
```
laurentlp, line 34: Remove this line and put the for loop inside the try-catch?!

packages/server/src/global.ts
```diff
@@ -12,7 +12,6 @@ export interface MessagingEnv {
   REDIS_SCOPE?: string
   LOGGING_ENABLED?: string
   SYNC?: string
-  SKIP_LOAD_CONFIG?: string
```
laurentlp, line 15: You should also remove this entry from the configs documentation.

packages/server/src/sync/service.ts
samuelmasse: Good point

packages/server/src/database/service.ts
samuelmasse: I prefer having the !

laurentlp (approved):

PR #154, opened 2021-08-24
fix(post): log errors as warnings

samuelmasse:
I my opinion webhook "errors" are not really server errors. So I changed the log status to be warnings.

Closes MES-144


laurentlp (approved):

samuelmasse:
Now looks like this
Screenshot from 2021-08-24 16-41-02

PR #158, opened 2021-08-24
fix(monitoring): always setup non lazy when spinned

samuelmasse:
Fixes a problem when spinned from botpress that would stop retrying to setup channels after having failed for 5 times (and the config staying the same).

### DIFF
--- a/packages/server/src/instances/monitoring.ts
+++ b/packages/server/src/instances/monitoring.ts
@@ -41,7 +41,7 @@ export class InstanceMonitoring {
 
     for (const outdated of outdateds) {
       const failures = await this.status.getNumberOfErrors(outdated.id)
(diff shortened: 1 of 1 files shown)


laurentlp:
> Fixes a problem when spinned from botpress that would stop retrying to setup channels after having failed for 5 times (and the config staying the same).

So the solution and to keep retrying forever?

samuelmasse:
Yes that's not a problem with botpress. It's just a problem in the cloud to not waste ressources.

samuelmasse:
Otherwise you could fail to setup your channel correctly on botpress for whatever reason (maybe ngrok is not started), and then you would have no idea why it doesn't work. The only way to make it work would be to modify the config of your channel to some alternate value to force messaging to try setting it up again

PR #159, opened 2021-08-25
chore(logging): added way to disable logging timestamp + prfixed logs with Redis node ID

laurentlp:
This PR adds a `DISABLE_LOGGING_TIMESTAMP` env var that can be used to disable logging timestamp (useful when logs are aggregated and timestamps added automatically by the aggregator). It also contains an improvement where the Redis node ID is prefixed when Messaging runs in cluster mode

e.g. 
Screenshot from 2021-08-25 16-18-25

Closes MES-139


samuelmasse (approved):
I pushed some fixes to correctly align the logger.window stuff for all possible configurations.

PR #160, opened 2021-08-30
chore(docker): added auto-migrate parameter to dockerfile start cmd

laurentlp:
Self-explanatory


samuelmasse (approved):
Do we do this in the botpress dockerfile?

laurentlp:
> Do we do this in the botpress dockerfile?

No, the alternative is to override the command when running messaging using the Dockerfile and then provide the value if you want it. I don't really have a preference, but if we consider that this Dockerfile will be used to upload an image of the Messaging server on Docker Hub, we should not go that way!

PR #168, opened 2021-09-02
feat(webhook): add convo and user events

samuelmasse:
This PR adds a `conversation created` and `user created` webhook event. I figured the way we were doing webhooks was all over the place so I created a StreamService who's role is to stream what happens on the server to listeners. These listeners can be sockets or webhooks. I also completely removed the chat service and api.

Closes DEV-1598

Related to https://github.com/botpress/botpress/pull/5406


packages/server/src/messages/api.ts
```diff
@@ -169,7 +167,7 @@ export class MessageApi extends ClientScopedApi {
         })
       }
 
-      const msg = await this.chat.send(conversationId, userId, payload, { socket })
+      const msg = await this.messages.create(conversationId, userId, payload) //, { socket })
```
samuelmasse, line 170: Chat api does not exist anymore. We'll have to add a `suppressEvents` options or something here to allow the channel web to only create messages without everything around.

samuelmasse: Turns out we actually don't need to do this

packages/server/src/stream/service.ts
```diff
@@ -0,0 +1,132 @@
...
+    this.health.events.on(HealthEvents.Registered, async ({ event }) => {
+      const conduit = await this.conduits.get(event.conduitId)
+      const client = await this.clients.getByProviderId(conduit!.providerId)
+      if (!client) {
+        return
+      }
+
+      const channel = this.channels.getById(conduit!.channelId)
+      await this.stream(
+        'health.new',
```
laurentlp, line 52: I think it would be nice if those events were put in a certain file (client maybe?) so that a user that listens to those events can easily register and does not have to read the code to find the list of available events. What do you think?

packages/server/src/stream/service.ts
```diff
@@ -0,0 +1,132 @@
...
+        conversation!.clientId,
+        conversation!.userId,
+        source
+      )
+    })
+  }
+
+  private async getChannel(conversationId: uuid) {
+    const convmaps = await this.mapping.convmap.listByConversationId(conversationId)
+    if (convmaps.length === 1) {
```
laurentlp, line 87: Again, not sure I understand this logic here.

packages/server/src/stream/service.ts
```diff
@@ -0,0 +1,132 @@
...
+      const tunnel = await this.mapping.tunnels.get(convmaps[0].tunnelId)
+      return this.channels.getById(tunnel!.channelId).name
+    } else {
+      return 'messaging'
+    }
+  }
+
+  async stream(type: string, payload: any, clientId: uuid, userId: uuid | undefined, source?: ActionSource) {
```
laurentlp, line 95:
```suggestion
  async stream(type: string, payload: any, clientId: uuid, userId?: uuid, source?: ActionSource) {
```

You can then remove the undefined on line 55

packages/server/src/stream/service.ts
```diff
@@ -0,0 +1,132 @@
...
+      for (const socket of sockets) {
+        if (source?.socket?.id !== socket.id) {
+          socket.send({ type, data })
+        }
+      }
+    }
+
+    // TODO: Send socket messages to sockets connected for a clientId
+
+    if (source?.client?.id !== clientId) {
```
laurentlp, line 120: Here also.

packages/server/src/stream/service.ts
samuelmasse: Yes we could have something similar to telegraf. We provide a middleware to be put in a router 

```ts
this.router.post(
  '/callback',
  messaging.callback()
)
```

and then we can do 

```ts
// messaging is an instance of MessagingClient
messaging.on('health.new', (msg) => { /* do something */ })
messaging.on('message.new', (msg) => { /* do something */  })
```
Alternatively we could do something like that as well. It's not as simple but it scales better  :
```ts
messaging.health.events.on(HealthEvents.New, (msg) => { /* do something */  })
messaging.messages.events.on(MessageEvents.New, (msg) => { /* do something */  })
```

For the socket we could use this approach as well. It could also be useful to use an object oriented approach to the objects sent in the callback. For example :

```ts
messaging.messages.events.on(MessageEvents.New, (msg) => {
  msg.reply(`Hello ${msg.userId}!`)
})

// instead of
messaging.messages.events.on(MessageEvents.New, async (msg) => {
  await messaging.messages.create(msg.conversationId, { type: 'text', text: `Hello ${msg.userId}!`)
})
```

packages/server/src/stream/service.ts
samuelmasse: If a conversation is only linked to one channel, we put the channel as that connected channel. Otherwise the conversation is happening on multiple channels, so we can't give a specific channel so we put 'messaging' as the channel.

packages/server/src/stream/service.ts
samuelmasse: I added some documentation to ActionSource.

laurentlp (approved):
Missing some tests but other than that, LGTM! Nice work!!!

PR #171, opened 2021-09-10
chore(dev): automating binary creation for branches

allardy:


samuelmasse:
What is this

allardy:
I described this in my PR on the main repository here; 
https://github.com/botpress/botpress/pull/5434

Basically, when you push a commit to a branch, github actions will create a binary of messaging (exactly the same thing as when you release a new version of messaging, but scoped to a branch). 

Then, on the Botpress repository, you can add "devBranch" next to "version" in package.json, so if you checkout the branch in botpress, it will download the binary of the developer branch instead of the master version (so we can test it correctly without having to publish on master, then change the version)

.github/workflows/release.yml
```diff
@@ -10,54 +10,46 @@ jobs:
...
-        id: validate_semver
-        run: |
-          valid_latest_tag=$(node ./node_modules/semver/bin/semver.js ${{ steps.previous_tag.outputs.tag }})
-          valid_current_version=$(node ./node_modules/semver/bin/semver.js ${{ steps.extract_version.outputs.version }})
-          change_log=$(node -e "require('./scripts/changelog.js').build({ writeToFile: true }).then(cl => console.log(cl))")
 
-          # if not change log is truncated at first line
-          change_log="${change_log//'%'/'%25'}"
-          change_log="${change_log//$'\n'/'%0A'}"
-          change_log="${change_log//$'\r'/'%0D'}"
```
samuelmasse, line 38: Why was this changed?

.github/workflows/pre-release.yml
```diff
@@ -0,0 +1,55 @@
+name: Pre-Release
```
samuelmasse, line 1: This config creates a new pre-release for every branch that is not master?

.github/workflows/branch_deleted.yml
```diff
@@ -0,0 +1,21 @@
+name: Branch deleted
```
samuelmasse, line 1: Ok so here we delete the pre-release whenever a branch gets deleted.

samuelmasse (commented):
Can you explain the changes a bit? Are the changes in release.yml just a refactor?

.github/workflows/pre-release.yml
allardy: Yes

.github/workflows/branch_deleted.yml
allardy: Yup, auto cleanup

.github/workflows/release.yml
allardy: This code was copied here, in the studio and in messaging. I created a single action which extracts all the information we need and made the release process simpler. You can check the information here:

https://github.com/botpress/gh-actions/blob/master/get_release_details/src/index.ts

samuelmasse (approved):

PR #172, opened 2021-09-11
chore(dev): add build info

allardy:
Standardizing build info

image


samuelmasse (approved):
I changed the implementation a bit to use logger.window

PR #176, opened 2021-09-14
fix(vonage): fix too many requests when using sandbox

laurentlp:
Description coming soon...


packages/server/src/channels/vonage/senders/common.ts
```diff
@@ -24,7 +24,7 @@ export class VonageCommonSender extends CommonSender {
                 }
 
                 context.logger.error(undefined, `${errBody.title}: ${errBody.detail} ${reasons}${errBody.type || ''}`)
-              } else if ((<any>err).statusCode === '429') {
+              } else if ((<any>err).statusCode === '429' || (<any>err).statusCode === 429) {
```
slvnperron, line 27: FYI you could just remove one `=`
```
> '429' == 429
true
```

PR #180, opened 2021-09-23
chore(dev): updated nodejs to v12.18.1

laurentlp:
This PR bumps the version of NodeJS to 12.18.1. Our goal is to be using this version across most of the repositories.


tomkcey:
Can't we bring it up to 14?

laurentlp:
> Can't we bring it up to 14?

I think we can even use the latest version of NodeJS, but we'll update to v14 when all the other repos are ready to do the same!

PR #194, opened 2021-10-13
feat(converse): converse api

samuelmasse:
Adds a collector to the message api to replicate behavior similar to the converse api

Closes DEV-1909
Closes DEV-1612

Requires to be on this branch when testing: https://github.com/botpress/botpress/pull/5568


allardy:
@samuelmasse The runtime will not be exposed to the outside world, so communications must go through messaging. Is this a good replacement for the converse api, and can we use it to benchmark messaging with botpress efficiently ?

samuelmasse:
Merging this right now. I'll make the final improvements in another PR

PR #205, opened 2021-10-15
chore(auth): refact auth

samuelmasse:
Extracted from #203. Does some refactoring for how we do authentification


samuelmasse:
Bugged actions closing

PR #206, opened 2021-10-15
chore(dx): update dev bins

allardy:
Since the dev bins was updated on studio/bp, i'm making the same change here so the "devBins" prop can be used in package.json of the main repo

It doesn't pollute the "releases" page, and i'll do a manual cleanup of the previous stuff


samuelmasse:
@allardy  We're getting some problems with the tags. Can you finish the fix you were doing here?

PR #211, opened 2021-10-18
chore(client): re-add internal passord and proxy support

laurentlp:
This PR simply re-adds support for passing an internal password to the messaging client and fixes an issue with proxies.


samuelmasse:
Most likely none of the changes here are necessary and the same results can be achieved by passing a custom axios client to the client parameter in the constructor

PR #221, opened 2021-10-20
feat(twilio): twilio testing

samuelmasse:
Adds a testing mode for twilio to tests calls to channels


samuelmasse:
That should do it

Screenshot from 2021-10-20 18-47-22

PR #223, opened 2021-10-26
feat(api): refact implementation

samuelmasse:
The code for the api was becoming a mess. Since this code does critical work for safety I needs to be refactored to be maintainable.

Todo: 
- [x] Update documentation for changed routes
- [ ] Open PR in core and runtime that updates the messaging version
- [x] Update message deletion function

Closes DEV-1945


packages/client/jest.config.ts
```diff
@@ -0,0 +1,19 @@
+import type { Config } from '@jest/types'
+import { defaults as tsjPreset } from 'ts-jest/presets'
+
+const config: Config.InitialOptions = {
+  preset: 'ts-jest',
+  projects: [
+    {
```
samuelmasse, line 7: I added tests for the client so we can tests our routes. It's a sort of semi-automatic testing like the socket tests. You need to start a server on port 3100 before running the tests

api.rest
```diff
@@ -65,14 +65,30 @@ x-bp-messaging-client-token: {{clientToken}}
...
+
+{
+  "conversationId": "{{convoId}}",
+  "authorId": "{{userId}}",
+  "payload": {
+    "type": "text",
+    "text": "Hello this is a converse message!"
+  }
+}
+
```
samuelmasse, line 91: This is the new route for converse. It's not a parameter in the query anymore since the behavior was way too different.

packages/server/src/conversations/api.ts
```diff
@@ -1,83 +1,54 @@
...
-        const conversation = await this.conversations.create(req.client!.id, userId)
-
-        res.send(conversation)
-      })
-    )
+  setup(router: ApiManager) {
+    router.post('/conversations', Schema.Api.Create, this.create.bind(this))
+    router.get('/conversations/:id', Schema.Api.Get, this.get.bind(this))
+    router.get('/conversations/user/:userId', Schema.Api.List, this.list.bind(this))
+    router.get('/conversations/user/:userId/recent', Schema.Api.Recent, this.recent.bind(this))
```
samuelmasse, line 15: The list messages by conversation route was changed from GET /conversations/:userId to GET /conversations/user/:userId. Basically the reasoning behind this is that we are listing all the `/conversations` by their `/user` property with the value `/:userId`. The same reasoning is applied the all other routes

samuelmasse: The more "object oriented" route `GET /user/:userId/conversations` isn't the name of the route here since it doesn't fit with the way messaging is made. In the messaging codebase services and object have no notion of where they are used. So for example the function to list the conversations of a user is in the ConversationService, not the UserService. It's made in a more bottom-up manner as opposed to top-down. So it makes sense that our routes also follow this logic. `GET /users/:userId/conversations` is the top-down way of doing it while `GET /conversations/user/:userId` is the bottom up way of doing it. @laurentlp let me know what you think

packages/server/src/base/schema.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export const ReqSchema = (
+  schema?: { body?: SchemaObject; params?: SchemaObject; query?: SchemaObject } | undefined
+) => {
+  return Joi.object({
+    body: Joi.object(schema?.body || {}).required(),
+    params: Joi.object(schema?.params || {}).required(),
+    query: Joi.object(schema?.query || {}).required()
+  }).required()
+}
```
samuelmasse, line 15: This is the new function to define request schemas. It always validate the body, params, and query parameters.

packages/client/src/messages.ts
```diff
@@ -1,39 +1,43 @@
...
-    }, [])
+  async delete(id: uuid): Promise<boolean> {
+    return (await this.http.delete<boolean>(`/messages/${id}`)).data
   }
 
-  async delete(filters: { id?: string; conversationId?: string }): Promise<number> {
-    return (await this.http.delete<{ count: number }>('/messages', { params: filters })).data.count
+  async deleteByConversation(conversationId: uuid): Promise<number> {
+    return (await this.http.delete<{ count: number }>(`/messages/conversation/${conversationId}`)).data.count
   }
```
samuelmasse, line 38: Notice here I split the message deletion routes into two. So we don't use query params anymore.

samuelmasse (commented):
@laurentlp I think this is good to merge now. There'll be 1 or 2 other PRs mostly for code refactors but the routes renaming should all be in this PR.

packages/server/src/conversations/api.ts
laurentlp: Doing `GET /user/:userId/conversations` is a more REST/RESTful way of doing things, but I understand that in our context `GET /conversations/user/:userId` makes more sense.

packages/server/src/messages/api.ts
```diff
@@ -1,178 +1,156 @@
...
+    const message = await this.messages.create(conversationId, authorId, payload, undefined, messageId)
 
-    this.router.post(
-      '/messages/turn/:id',
-      this.auth.client.auth(async (req, res) => {
-        const { error } = Schema.Api.Turn.validate(req.params)
-        if (error) {
-          return res.status(400).send(error.message)
-        }
+    res.send({ message, responses: await collector })
```
laurentlp, line 77:
```suggestion
    res.status(201).send({ message, responses: await collector })
```

packages/server/src/messages/api.ts
```diff
@@ -1,178 +1,156 @@
...
-        if (error) {
-          return res.status(400).send(error.message)
-        }
+    const conversation = await this.conversations.get(message.conversationId)
+    if (!conversation || conversation.clientId !== req.client.id) {
+      return res.send(false)
+    }
+
+    await this.messages.delete(id)
+    res.send(true)
```
laurentlp, line 123: Usually with delete call, a 204 no content means that the resource was properly deleted.

packages/server/src/messages/api.ts
```diff
@@ -1,178 +1,156 @@
...
-        const messages = await this.messages.listByConversationId(conversationId as uuid, +(limit as string))
+    const deleted = await this.messages.deleteByConversationId(conversationId)
+    res.send({ count: deleted })
+  }
 
-        res.send(messages)
-      })
-    )
+  async turn(req: ClientApiRequest, res: Response) {
+    const { id } = req.params
```
laurentlp, line 139: No casting here?

packages/server/src/messages/schema.ts
```diff
@@ -1,44 +1,69 @@
...
   }),
 
-  Turn: Joi.object({
-    id: Joi.string().guid().required()
+  Collect: ReqSchema({
+    body: {
+      conversationId: Joi.string().guid().required(),
+      authorId: Joi.string().guid().required(),
+      payload: Joi.object().required(),
+      timeout: Joi.number().min(0).optional()
```
laurentlp, line 19: We should add a max value as well.

packages/server/src/messages/schema.ts
```diff
@@ -1,44 +1,69 @@
...
 
-  List: Joi.object({
-    conversationId: Joi.string().guid().required(),
-    limit: Joi.number().required()
+  List: ReqSchema({
+    params: {
+      conversationId: Joi.string().guid().required()
+    },
+    query: {
+      limit: Joi.number().optional()
```
laurentlp, line 34:
```suggestion
      limit: Joi.number().min(0).optional()
```

packages/server/src/sync/api.ts
```diff
@@ -9,53 +10,38 @@ import { makeSyncRequestSchema } from './schema'
...
-      })
-    )
+  private force!: boolean
+  private schema!: Joi.ObjectSchema
+
+  constructor(private syncs: SyncService, private clients: ClientService, private channels: ChannelService) {}
+
+  setup(router: Router, auth: Auth) {
+    this.force = !!yn(process.env.SPINNED)
+    this.schema = makeSyncRequestSchema(this.channels.list())
```
laurentlp, line 20: Why not put this in the constructor?

packages/server/src/conversations/api.ts
```diff
@@ -1,83 +1,54 @@
...
-        const { userId } = req.body
-        const conversation = await this.conversations.create(req.client!.id, userId)
-
-        res.send(conversation)
-      })
-    )
+  setup(router: ApiManager) {
+    router.post('/conversations', Schema.Api.Create, this.create.bind(this))
+    router.get('/conversations/:id', Schema.Api.Get, this.get.bind(this))
+    router.get('/conversations/user/:userId', Schema.Api.List, this.list.bind(this))
```
laurentlp, line 14: Should be:
```suggestion
    router.get('/conversations/users/:userId', Schema.Api.List, this.list.bind(this))
```

packages/server/src/base/api-manager.ts
```diff
@@ -0,0 +1,35 @@
...
+    this.use('delete', path, schema, fn)
+  }
+
+  use(type: 'post' | 'get' | 'delete', path: string, schema: Joi.ObjectSchema<any>, fn: Middleware<ClientApiRequest>) {
+    this.router[type](
+      path,
+      this.auth.client.auth(async (req, res) => {
+        const { error } = schema.validate({ query: req.query, body: req.body, params: req.params })
+        if (error) {
+          return res.status(400).send(error.message)
```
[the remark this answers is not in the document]
samuelmasse: 422 makes sense but as is discussed in that post it seems to be an extension code. I've personally never seen this code anywhere so I'd say a 400 is better simply because it's the most used. We're sending the full error message here anyway so I doubt this response would be confusing to anyone.

packages/server/src/base/api-manager.ts
```diff
@@ -0,0 +1,35 @@
...
+        }
+
+        await fn(req, res, <any>undefined)
```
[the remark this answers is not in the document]
samuelmasse: We don't use it anywhere. In fact now that you mention it I removed it from the accepted parameters so we don't need to send this dummy undefined anymore.

packages/server/src/conversations/api.ts
```diff
@@ -1,83 +1,54 @@
...
-        const { error } = Schema.Api.Get.validate(req.params)
-        if (error) {
-          return res.status(400).send(error.message)
-        }
+  async create(req: ClientApiRequest, res: Response) {
+    const userId = req.body.userId as uuid
 
-        const { id } = req.params
-        const conversation = await this.conversations.get(id)
+    const conversation = await this.conversations.create(req.client!.id, userId)
```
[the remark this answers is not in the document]
samuelmasse: I believe this could be undefined before so there's some crap left.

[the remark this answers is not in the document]
samuelmasse: This is implicit in the listByUserId function of the service. We query the db with WHERE userId = and clientId = so we just don't get results that don't belong to the user and client.

packages/server/src/conversations/api.ts
```diff
@@ -1,83 +1,54 @@
...
-        }
+  async get(req: ClientApiRequest, res: Response) {
+    const id = req.params.id as uuid
 
-        res.send(conversation)
-      })
-    )
+    const conversation = await this.conversations.get(id)
+    if (!conversation || conversation.clientId !== req.client!.id) {
+      return res.send(undefined)
```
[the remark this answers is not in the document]
samuelmasse: We return undefined if `no value` is a correct value to get. 404 implies a client error. Since our client indicates that the get function can return a value or undefined, then undefined should be a value that the server can return. The get function is both a "get properties of that object" and "check if that object exists"

packages/server/src/conversations/api.ts
```diff
@@ -1,83 +1,54 @@
...
+    const userId = req.params.userId as uuid
 
-        const { userId } = req.params
-        let conversation = await this.conversations.getMostRecent(req.client!.id, userId)
-        if (!conversation) {
-          conversation = await this.conversations.create(req.client!.id, userId)
-        }
+    let conversation = await this.conversations.getMostRecent(req.client!.id, userId)
+    if (!conversation) {
+      conversation = await this.conversations.create(req.client!.id, userId)
```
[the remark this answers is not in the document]
samuelmasse: This doesn't change anything from the previous behavior of the api. This route was made like that to work with the channel-web backend. We might completely remove it once we convert the webchat.

packages/server/src/conversations/api.ts
```diff
@@ -1,83 +1,54 @@
...
-          +(limit as string)
-        )
+  async list(req: ClientApiRequest, res: Response) {
+    const userId = req.params.userId as uuid
+    const limit = +(req.query.limit || 20)
 
-        res.send(conversations)
-      })
-    )
+    const conversations = await this.conversations.listByUserId(req.client!.id, userId, limit)
```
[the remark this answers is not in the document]
samuelmasse: While mismatching the userId and clientId would return an empty list, I agree that trying to list the convos of a user that doesn't belong to the client should return 403.

packages/server/src/messages/api.ts
```diff
@@ -1,178 +1,156 @@
...
+    router.delete(
+      '/messages/conversation/:conversationId',
+      Schema.Api.DeleteByConversation,
+      this.deleteByConversation.bind(this)
+    )
+    router.post('/messages/turn/:id', Schema.Api.Turn, this.turn.bind(this))
+  }
+
+  async create(req: ClientApiRequest, res: Response) {
+    const { conversationId, authorId, payload, incomingId } = req.body as {
```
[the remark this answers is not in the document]
samuelmasse: Will do in next PR

packages/server/src/messages/api.ts
```diff
@@ -1,178 +1,156 @@
...
+      conversationId: uuid
+      authorId: uuid
+      payload: any
+      incomingId: uuid
+    }
+
+    const conversation = await this.conversations.get(conversationId)
+    if (!conversation || conversation.clientId !== req.client.id) {
+      return res.sendStatus(404)
```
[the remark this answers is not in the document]
samuelmasse: We return 404 if you're trying to use a ressource that either doesn't exist or doesn't belong to you. We return undefined if you're fetching a single ressource. So the only case where we return undefined instead of 404 is in the get function because it's valid for the api to lookup if a ressource exist that way. On the other hand, trying to create a message for a conversation that doesn't exist can't return anything else than an error because it's simply wrong api usage.

packages/server/src/messages/api.ts
samuelmasse: Good point forgot that

packages/server/src/messages/api.ts
samuelmasse: For this route we consider it valid to try to delete a ressource that doesn't exist. What should we return when the ressource doesn't exist? If we returned an error code it would mean that the route doesn't accept trying to delete something that is already deleted.

packages/server/src/sync/api.ts
samuelmasse: It would work put in theory this function relies on data from another service which should only be used after that service was setup.

packages/server/src/conversations/api.ts
samuelmasse: I don't think so. We're listing `conversations` by their `user` property. I don't see the need for plural here.

packages/server/src/base/api-manager.ts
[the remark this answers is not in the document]
laurentlp: Alright sounds good.

packages/server/src/conversations/api.ts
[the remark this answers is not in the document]
laurentlp: In this case, it is hard to know what caused the 404 error. Is it because the user does not exist or because it is not associated with the client.

packages/server/src/conversations/api.ts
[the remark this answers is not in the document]
laurentlp: Taken from Wikipedia:

| ..., to indicate that the browser was able to communicate with a given server, but the server could not find what was requested. The error may also be used when a server does not wish to disclose whether it has the requested information.

I don't see how this is a client error.

If we were to implement a client in another language, `undefined` might not make sense. I think we should use standard HTTP code as much as possible so we are not faced with surprises in the future. The more we are close to standard the easier it is to connect with any other standard tool.

packages/server/src/conversations/api.ts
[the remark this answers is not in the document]
laurentlp: Fair enough. Can you add a TODO to remove this block of code once we migrate the channel-web as I think it is too custom to be part of the API.

packages/server/src/messages/api.ts
laurentlp: I might be mistaken, but if I recall correctly, the standard is to always return 204 no matter what (maybe 404 too, not sure).

Look at the second answer from this thread, I think it makes sense: https://stackoverflow.com/questions/6439416/status-code-when-deleting-a-resource-using-http-delete-for-the-second-time

packages/server/src/conversations/api.ts
laurentlp: Usually entities in REST routes are all plural no matter what.

packages/server/src/conversations/api.ts
```diff
@@ -35,18 +43,23 @@ export class ConversationApi {
 
   async list(req: ClientApiRequest, res: Response) {
     const userId = req.params.userId as uuid
-    const limit = +(req.query.limit || 20)
+    const limit = +(req.query.limit || DEFAULT_LIMIT)
+
+    const user = await this.users.get(userId)
+    if (!user || user.clientId !== req.client.id) {
```
laurentlp, line 49: I know we already had this discussion, but can you remind me why if you do a request to list the conversations of a user that is linked with a different client than yours, you would get a 404 not found instead of a 403 Forbidden?

In my mind, if you try to access a resource that you don't have permission to. Be it an admin page or user information, it should return 403, not 404.

Let's say I am a not user with normal privileges on a given website. If I were to access the admin page, I expect the error code to be 403 and then to redirect me back to where I was.

samuelmasse: Messaging is multi tenant for performance purposes, but we divide the database in different sectors to manage ressource ownership. The clientIds split the database into multiple smaller virtual databases if you want. So it makes sense in that context that trying to access a resource that's located outside your "virtual database" would return not found, since in your point of view the item is not located in your database.

laurentlp: Fair point! As long as we are consistent here, let's go that way! And also, this makes me think that we should document that as it isn't that obvious!

packages/server/src/conversations/api.ts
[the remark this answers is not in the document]
samuelmasse: As explained in my other comment, from the clientId's "sandbox" a item existing somewhere else or not at all is the same thing. So it just doesn't exist

packages/server/src/conversations/api.ts
```diff
@@ -1,83 +1,54 @@
...
 
-    this.router.get(
-      '/conversations/:id',
-      this.auth.client.auth(async (req, res) => {
-        const { error } = Schema.Api.Get.validate(req.params)
-        if (error) {
-          return res.status(400).send(error.message)
-        }
+  async create(req: ClientApiRequest, res: Response) {
+    const userId = req.body.userId as uuid
```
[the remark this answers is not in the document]
samuelmasse: We'll check if something like that is possible in another PR

packages/server/src/conversations/api.ts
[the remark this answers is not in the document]
samuelmasse: 404 is an **error** message. Trying to fetch an item that does not exist is not defined as an error by the api, because we want this route to be useable to check if an item exists. We return undefined as a null value. If we were to adapt the client to any other language we would just use whatever version of `null` that language supports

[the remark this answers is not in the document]
laurentlp: It shouldn't be the API "job" to define what is an error or not. Since we are defining and standardizing a REST (HTTP) API, it makes sense to stay close to the [RFCs](https://www.rfc-editor.org/rfc/rfc7231#section-6.5.4) whose goal is to define how the API should behave. 404 means undefined in HTTP. An HTTP 200 code means that the server found the value, that the request has succeeded. When we return undefined here, the request has not succeeded in finding the requested value.

I understand your point of view that if the request returns nothing, we should do the same thing, but doing so is not the norm. I don't see any disadvantages of returning 404 like all the other API we might use will do.

Furthermore, since we are going to mainly use our clients, they simply have to handle this abstraction of switching 404s by undefined.

It's your call, but I warn you that the more we distance ourselves from the standard, the more we are prone to have unexpected issues in the future.

laurentlp (approved):
Solid work here, as usual!!!! :100:

PR #227, opened 2021-11-01
fix(messenger): fix fetch pageId when in live mode

samuelmasse:
Fixes a problem where having a bot in live mode wouldn't work if it wasn't approved by facebook.

Closes DEV-1950


franklevasseur (approved):
LGTM

PR #229, opened 2021-11-01
fix(channels): fixes for smooch and messenger

franklevasseur:
Head branch of team shake_your_botpress

Members of the team are: 
- @franklevasseur 
- @samuelmasse 
- @daukadolt 
- @crochcunill

Bug fixes:

- fix(messenger): fix fetch pageId when in live mode (**2pts**)
- fix(smooch): fix buttons (**2pts**)

**DO NOT SQUASH THIS BRANCH**

### DIFF
--- a/packages/server/src/channels/smooch/conduit.ts
+++ b/packages/server/src/channels/smooch/conduit.ts
@@ -6,7 +6,7 @@ import { ChannelContext } from '../base/context'
 import { CardToCarouselRenderer } from '../base/renderers/card'
 import { DropdownToChoicesRenderer } from '../base/renderers/dropdown'
 import { SmoochConfig } from './config'
-import { SmoochMessage, SmoochPayload, SmoochContext, SmoochWebhook } from './context'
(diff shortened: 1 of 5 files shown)


packages/server/src/channels/smooch/conduit.ts
```diff
@@ -62,9 +62,24 @@ export class SmoochConduit extends ConduitInstance<SmoochConfig, SmoochContext>
...
-  public async extractEndpoint(payload: { context: SmoochPayload; message: SmoochMessage }): Promise<EndpointContent> {
+  public async extractEndpoint(payload: {
+    context: SmoochPayload
+    message: SmoochMessage
+    action: SmoochAction
+  }): Promise<EndpointContent> {
+    const postback = payload.message.action?.payload
+    let content
+
+    if (postback?.startsWith('say::')) {
```
laurentlp, line 73: `say::` and `postback::` should be moved into constants somewhere. If they could be used in other channels, maybe put them in the base channel.

laurentlp (commented):
Small suggestions otherwise, looks good to me! Nice work

samuelmasse (approved):

PR #230, opened 2021-11-01
fix(smooch): fix buttons

samuelmasse:
Closes DEV-1625

Fixes postback and say something buttons on smooch


franklevasseur (approved):
LGTM!

PR #232, opened 2021-11-01
chore(repo): unified test config

EFF:


samuelmasse:
So this merges all configs for jest in a single file. Is this a bug fix?

laurentlp:
> So this merges all configs for jest in a single file. Is this a bug fix?

No only dx improvements!

samuelmasse (approved):

PR #235, opened 2021-11-03
chore(webchat): display something

samuelmasse:
WIP

Starts rendering the webchat in the browser using the board package.

Closes DEV-1884


packages/webchat/src/components/messages/renderer/Carousel.tsx
```diff
@@ -2,8 +2,9 @@ import React from 'react'
 import Slider from 'react-slick'
 
 // Added those manually to remove the font dependencies which keeps showing 404 not found
-import '../../../../../../assets/slick/slick-theme.css'
-import '../../../../../../assets/slick/slick.css'
+// TODO: see what to do with these
+// import '../../../../../../assets/slick/slick-theme.css'
+// import '../../../../../../assets/slick/slick.css'
```
allardy, line 7: This will go on a CDN. Keep all the files from /assets in a folder in the repository, we'll need them when running the webchat locally and we'll just push the whole assets folder on the CDN

packages/webchat/src/index.tsx
```diff
@@ -1,4 +1,5 @@
-import '@blueprintjs/core/lib/css/blueprint.css'
+// TODO: figure out what to do with this
+// import '@blueprintjs/core/lib/css/blueprint.css'
```
allardy, line 2: cdn also

packages/webchat/src/main.tsx
```diff
@@ -34,17 +34,23 @@ class Web extends React.Component<MainProps> {
   }
 
   async componentDidMount() {
+    // TODO: 1 - we need to get this from somewhere else. CDN of audio file?
```
allardy, line 37: Yes, cdn will serve all assets

packages/webchat/src/main.tsx
```diff
@@ -34,17 +34,23 @@ class Web extends React.Component<MainProps> {
...
     this.audio = new Audio(`${window.ROOT_PATH}/assets/modules/channel-web/notification.mp3`)
+    // TODO: 2 - can we remove intl?
     this.props.store.setIntlProvider(this.props.intl!)
```
allardy, line 40: No, we need translations

packages/webchat/src/main.tsx
```diff
@@ -34,17 +34,23 @@ class Web extends React.Component<MainProps> {
...
+
+    // TODO: 3 - we migth want to watch how much we use global variables with window. Why is this globally available
     window.store = this.props.store
 
+    // TODO: 4 - iframe api? Why do we need that?
```
allardy, line 45: The webchat runs in an iframe. It will still be the case. (mostly for styling issues).

packages/webchat/src/main.tsx
```diff
@@ -34,17 +34,23 @@ class Web extends React.Component<MainProps> {
...
     window.addEventListener('message', this.handleIframeApi)
     window.addEventListener('keydown', (e) => {
       if (!this.props.config!.closeOnEscape) {
         return
       }
       if (e.key === 'Escape') {
         this.props.hideChat!()
+        // TODO: 5 - what to do with emulator mode?
         if (this.props.config!.isEmulator) {
           window.parent.document!.getElementById('mainLayout')!.focus()
```
allardy, line 55: This was added because keyboard shortcuts were broken when you used "esc", the focus was still on the webchat so no other shortcuts were working until you clicked somewhere else

packages/webchat/src/main.tsx
```diff
@@ -76,6 +82,7 @@ class Web extends React.Component<MainProps> {
     if (this.props.activeView === 'side' || this.props.isFullscreen) {
       this.hasBeenInitialized = true
 
+      // TODO: 17 - isn't this duplicated logic? Why are we doing this again?
```
allardy, line 85: Somehow, it fixed something

packages/webchat/src/main.tsx
```diff
@@ -88,19 +95,30 @@ class Web extends React.Component<MainProps> {
   async load() {
     this.config = this.extractConfig()
 
+    // TODO: 6 - this implies iframe as well. Do we need this? Why?
     if (this.config.exposeStore) {
       const storePath = this.config.chatId ? `${this.config.chatId}.webchat_store` : 'webchat_store'
       set(window.parent, storePath, this.props.store)
     }
```
allardy, line 102: I think we could remove it, because exposing the store doesnt work across domains (iframe vs parent frame). But lets leave it for now.

packages/webchat/src/main.tsx
```diff
@@ -88,19 +95,30 @@ class Web extends React.Component<MainProps> {
...
 
-    this.config.overrides && this.loadOverrides(this.config.overrides)
+    if (this.config.overrides) {
+      this.loadOverrides(this.config.overrides)
+    }
 
-    this.config.containerWidth && this.postMessageToParent('setWidth', this.config.containerWidth)
+    // TODO: 7 - again with the iframe stuff. Why do we need this.
+    if (this.config.containerWidth) {
```
allardy, line 109: The studio will still load the iframe of the webchat. Each websites will load the chat in the iframe.

packages/webchat/src/main.tsx
```diff
@@ -88,19 +95,30 @@ class Web extends React.Component<MainProps> {
...
+      this.postMessageToParent('setWidth', this.config.containerWidth)
+    }
 
-    this.config.reference && this.props.setReference!()
+    // TODO: 8 - what is reference?
+    if (this.config.reference) {
+      await this.props.setReference!()
```
allardy, line 115: Reference is when you need to pass informations in the URL which should not be altered. The reference is a signature that certifies the payload

packages/webchat/src/main.tsx
```diff
@@ -88,19 +95,30 @@ class Web extends React.Component<MainProps> {
...
+    }
 
+    // TODO: 9 - don't think we need to fetch bot info from a backend. All this can be set at the frontend level
```
allardy, line 118: Aren't we on the frontend? Also we must not forget that the runtime will not be directly reachable, so the frontend cant directly call it to fetch the informations

packages/webchat/src/main.tsx
```diff
@@ -88,19 +95,30 @@ class Web extends React.Component<MainProps> {
...
     await this.props.fetchBotInfo!()
 
+    // TODO: 10 - That's something we should still have as configurable. We just don't start the socket connection right away
     if (!this.isLazySocket()) {
       await this.initializeSocket()
```
allardy, line 123: Yeah, because we want to initialize the webchat once the user clicks on the webchat icon on the bottom right corner. It's useless to initialize before an interaction because it just causes a ton of usage

packages/webchat/src/main.tsx
```diff
@@ -120,11 +138,12 @@ class Web extends React.Component<MainProps> {
         return options
       }
     }
+    // TODO: 14 - parsing the config from the url is something that we don't need to do ourselves imho
```
allardy, line 141: The shortlink /s/welcome-bot was used to redirect to the bot URL and configure various options, but I agree that we could move all the configuration in the messaging channel's config of the webchat.

Basically the studio received that URL then it loaded the webchat and displayed it to the users. We'll still need the reference tho, maybe other stuff also. we'll have to discuss

packages/webchat/src/main.tsx
```diff
@@ -148,6 +168,7 @@ class Web extends React.Component<MainProps> {
   loadOverrides(overrides: Overrides) {
     try {
       for (const override of _values(overrides)) {
+        // TODO: 15 - load module view this can't work
         override.map(({ module }) => this.props.bp!.loadModuleView(module, true))
```
allardy, line 172: We will need to keep the load view logic. Custom components in the chat will use them, and we'll probably also keep the override mechanism for components, they will be bundled with the bot and served from the cdn

packages/webchat/src/main.tsx
```diff
@@ -168,12 +190,15 @@ class Web extends React.Component<MainProps> {
     })
 
     observe(this.props.config!, 'overrides', (data: any) => {
+      // TODO: 12 - we didn't check if window.parent was set the first time
+      // we loaded overrideds, so why do we do now?
       if (data.newValue && window.parent) {
         this.loadOverrides(data.newValue)
```
allardy, line 196: I think that was only if you changed the config on the emulator to see the change in real time

packages/webchat/src/main.tsx
```diff
@@ -168,12 +190,15 @@ class Web extends React.Component<MainProps> {
...
       }
     })
 
     observe(this.props.dimensions!, 'container', (data: any) => {
+      // TODO: 13 - the first time we do setWith we do it with props.containerWidth. Why the sudden change to props.dimensions?
       if (data.newValue && window.parent) {
         this.postMessageToParent('setWidth', data.newValue)
```
allardy, line 203: containerWidth / layoutWidth are the configurations you can change when initializing the webchat. Here we listen if the dimension of the container changes, if it does, we call the parent iframe to resize correctly so the webchat is displayed correctly.

Behind the scene, we store both values in props.dimensions (container/layout).

packages/webchat/src/utils.tsx
```diff
@@ -17,6 +17,7 @@ export const downloadFile = (name: string, blob: Blob) => {
   window.URL.revokeObjectURL(url)
 }
 
+// TODO: 18 - this function is likely not necessary anymore.
```
allardy, line 20: I dont remember why it was added. But it fixed something. Maybe older browser

allardy (commented):
I see a lot of questions about the frontend logic, please let's not question those choices and we need to focus on integrating the backend.

Each features on the webchat should stay the same. In the best scenario, you would only have to adjust the file `api.tsx` to adapt calls to the backend and adjust them.

samuelmasse:
Text is working so far

Screenshot from 2021-11-05 18-23-00

packages/webchat/src/main.tsx
samuelmasse: Alright then

packages/webchat/src/main.tsx
allardy: Technically the reference is just used to send a "signed" message to the backend. When you set a reference on the bot's URL, we validate the signature and if it's valid, we send an event of type "session_reference" with the valid payload.

For example, imagine you have a ticketing system and you receive e-mails everytime a new ticket is created. You have 3 different tickets: tic-1, tic-2, tic-3. 

The link in each email would open the webchat on the specific ticket. If we just add a query string to the url, for example `?ticketId=1`, someone could just forge the id and use a different ID. To avoid this, you can sign the payload using the sdk, so the reference might look like `?ticketId=1=de6f53gr6g59h5yr2v1df6f9eg`. This prevents the end user to alter the message.

packages/webchat/src/main.tsx
allardy: Ideally, we could avoid to put configurations in "window.botpressWebChat.configure" and messaging could provide the configurations from bot.config.json to the frontend directly

packages/webchat/src/main.tsx
allardy: I mean, the configuration of the channel in bot.config.json

packages/webchat/src/core/api.tsx
```diff
@@ -133,17 +94,20 @@ export default class WebchatApi {
...
+  // TODO: we don't have a /events route available for this
+  /*
   async sendEvent(payload: any, conversationId: uuid): Promise<void> {
     try {
       return this.axios.post('/events', { ...this.baseUserPayload, conversationId, payload }, this.axiosConfig)
     } catch (err) {
       await this.handleApiError(err)
     }
   }
+  */
```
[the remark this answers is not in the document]
allardy: This was used to send a payload other than "message" to the server (for example, proactive messages). sendMessage enforces the "text" payload for the incoming message. Can we use a custom payload type when sending a normal message ?

packages/webchat/src/core/api.tsx
```diff
@@ -231,10 +196,5 @@ export default class WebchatApi {
       console.error('Conversation not found, starting a new one...')
       await this.createConversation()
     }
-
-    if (data.errorCode === 'BP_0401') {
-      this.updateAxiosConfig({ botId: this.botId, externalAuthToken: undefined })
-      console.error('External token expired or invalid. Removed from future requests')
-    }
   }
```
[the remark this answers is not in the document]
allardy: When you configure the webchat, you can also set an "externalAuthToken" (a JWT token). Everytime you send a message, an event or some other requests, we validate that token and if it is valid, we add the property ".credentials" to the event when sending it to the event engine.

In essence, this is an authentication mechanism to compensate for the lack of security provided by the "visitorId". 

If the token is deemed invalid, we remove it from future requests to avoid parsing it each time

packages/webchat/src/core/api.tsx
```diff
@@ -205,6 +169,7 @@ export default class WebchatApi {
...
 
   async setReference(reference: string, conversationId: uuid): Promise<void> {
+    // TODO: this can't work. We don't have a reference route
     try {
       return this.axios.post(
         '/conversations/reference',
```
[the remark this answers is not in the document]
allardy: I've explained it more thoroughly in another comment, we will need this feature

packages/webchat/src/core/api.tsx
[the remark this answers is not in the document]
samuelmasse: We now have secure connection with the userId and userToken. Do we still need this feature?

packages/webchat/src/core/api.tsx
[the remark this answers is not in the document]
samuelmasse: You can use a custom payload when sending a message but that message will still be stored as a message in the db (as opposed to the /events route that from what I can see only created an event but not a message in the db). Maybe we could add the concept of transient message to messaging? So the message would be sent to webhooks but wouldn't be stored as a message. This same concept could be used for the user_visit

packages/webchat/src/core/api.tsx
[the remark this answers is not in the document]
allardy: The externalAuthToken also provides more informations, since you can put anything in that payload (and can be assured that the content of `event.credentials` is safe & valid). Securing with a id/token is a good start, but it doesn't really replace this feature.

@EFF @slvnperron i'm not entirely sure how it's being used by clients (or if it is being used). Your thoughts?

packages/webchat/src/core/api.tsx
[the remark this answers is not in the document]
allardy: Sounds good to me

packages/webchat/src/core/socket.tsx
```diff
@@ -1,73 +1,95 @@
...
     this.events.on('guest.webchat.clear', this.onClear)
     this.events.on('guest.webchat.message', this.onMessage)
     this.events.on('guest.webchat.typing', this.onTyping)
     this.events.on('guest.webchat.data', this.onData)
+    */
 
     // firehose events to parent page
-    this.events.onAny(this.postToParent)
+    // TODO: why do we need this
+    // this.events.onAny(this.postToParent)
```
allardy, line 32: This way you can add a custom handler in your page to handle any incoming events with `window.addEventListener('message')`

packages/webchat/src/core/socket.tsx
```diff
@@ -1,73 +1,95 @@
...
     })
+
+    return this.waitingForUser
   }
-}
 
-const isString = (str: string | any): str is string => {
-  return typeof str === 'string' && str !== 'undefined'
+  public getStorage<T>(key: string): T | undefined {
+    const stored = localStorage.getItem(this.getStorageKey(key))
```
allardy, line 75: Please copy the `storage.ts` file (from ui-shared-lite) instead of implementing a custom version. It supports localStorage, cookies and sessionStorage, and automatically serialize/deserialize infromations.

In some cases the local storage is disabled as a security measure.

There is also a configuration option to use sessionStorage instead of local

packages/webchat/src/store/index.ts
```diff
@@ -240,7 +245,8 @@ class RootStore {
 
   @action.bound
   async fetchPreferences(): Promise<void> {
-    const preferences = await this.api.fetchPreferences()
+    // TODO: where to fetch this from? Can this just be set in the frontend?
+    const preferences = { language: 'en' } // await this.api.fetchPreferences()
```
allardy, line 249: This comes from the user's attributes, and it affects the webchat UI and the language in which the bot will respond.  When you enable the bot's details page, you also have a drop down to change language. Just storing the lang in the frontend won't do it

packages/webchat/src/store/index.ts
```diff
@@ -252,16 +258,14 @@ class RootStore {
...
+    const conversations = await this.api.fetchConversations()
 
     runInAction('-> setConversations', () => {
-      if (!conversations.length) {
+      if (!conversations?.length) {
         this.view.showBotInfo()
       }
 
-      this.config.recentConversationLifetime = recentConversationLifetime
-      this.config.startNewConvoOnTimeout = startNewConvoOnTimeout
```
allardy, line 263: These two configurations are actually used often. There could be a single endpoint on messaging's side where you can get the bot's configuration, webchat config and these two configs

packages/webchat/src/index.tsx
```diff
@@ -46,6 +46,7 @@ class ExposedWebChat extends React.Component<Props, State> {
   }
 }
 
+// TODO: what does this observer do?
 const Wrapper = observer(ExposedWebChat)
```
samuelmasse, line 51: @allardy Any comments on this? I don't use the wrapper at the moment and I haven't see much difference.

allardy: When you have those kind of questions, you can hover on the line and it will display the related commit, giving a reason why this was added.
image

Since you hardcoded english on the fetch preferences, you might not have seen it

samuelmasse:
@laurentlp Let's keep in mind the issues in this PR. Now since this isn't published anywhere we'll merge this PR to master to be able to progress faster on other PRs that depend on these changes.

laurentlp (approved):

PR #236, opened 2021-11-05
chore(components): move component renderers to new typings

charlescatta:
In working condition, only small stuff left:

TODOS:

- [x] Validate dropdown typings with @samuelmasse (validated via msg_messages table payload)
- [x] Fix bug in dropdown not calling onSendData on select
- [x] Find a way to test this
- [x] Fix some types requiring the type: "typename" argument when using a content type as a subtype


packages/components/src/typings.d.ts
```diff
@@ -1,16 +1,13 @@
+import { Content, MessageType } from '@botpress/messaging-server/content-types'
 import { AxiosInstance } from 'axios'
 import { InjectedIntl } from 'react-intl'
-import { Settings as CarouselSettings } from 'react-slick'
```
samuelmasse, line 3: I don't think this package should depend on the server code! You can just copy paste the file from the server for now, and we'll probably move it to the `base` package later.

packages/components/package.json
```diff
@@ -26,7 +26,8 @@
     "build-storybook": "build-storybook"
   },
   "files": [
-    "dist"
+    "dist",
+    "src/typings.d.ts"
   ],
```
samuelmasse, line 31: If you're having trouble with this .d.ts file you might want to just rename it to `typings.ts`. `.d.ts` files are kind of weird to have in source code. For example the typescript compiler doesn't even copy them over to the output directory. If you rename it to `.ts` the compiler will produce a `.d.ts` for you anyways

charlescatta: alright!

packages/components/src/typings.d.ts
charlescatta: done

packages/components/package.json
charlescatta: done

charlescatta:
@samuelmasse I plan on testing this by pulling payloads from the event table from the database, good plan?

samuelmasse:
@charlescatta Yes this should work. You can also take payloads from the msg_messages table in the payload column

samuelmasse:
So this is VERY nitpicky but usually we don't use uppercase for our PR titles :sweat_smile:

packages/components/src/content-typings.ts
```diff
@@ -0,0 +1,162 @@
...
+  'quick_reply',
+  'login_prompt',
+  'session_reset',
+  'custom',
+  'unsupported' // This is a special type that is used as fallback for unsupported message types at runtime
+] as const
+
+type MessageTypeTuple = typeof messageTypes
+
+export type MessageType = MessageTypeTuple[number]
```
laurentlp, line 23: Wouldn't that work instead?

```suggestion
export type MessageType =
  | 'text'
  | 'audio'
  | 'file'
  | 'video'
  | 'image'
  | 'dropdown'
  | 'visit'
  | 'voice'
  | 'typing'
  | 'card'
  | 'carousel'
  | 'location'
  | 'quick_reply'
  | 'login_prompt'
  | 'session_reset'
  | 'custom'
  | 'unsupported' // This is a special type that is used as fallback for unsupported message types at runtime
```

packages/components/src/renderer/Dropdown/index.tsx
```diff
@@ -36,61 +35,73 @@ export const Dropdown = ({ payload, config }: MessageTypeHandlerProps<'dropdown'
...
+    if (buttonText && displayInKeyboard) {
+      return
+    }
+    try {
+      void (async () => {
+        await sendChoice()
+      })()
+    } catch (err) {
+      console.error(err)
+    }
```
laurentlp, line 48:
```suggestion
        sendChoice().catch(console.error)
```

packages/components/src/renderer/FileMessage/index.tsx
```diff
@@ -1,53 +1,52 @@
...
+  )
+}
 
+export const File: React.FC<MessageTypeHandlerProps<'file'>> = ({ file, title }) => {
   return (
     <div>
       <span>File: </span>
-      <a href={url} target={'_blank'}>
-        {title || url}
+      <a href={file} target={'_blank'}>
```
laurentlp, line 22: This should remove the warning.

```suggestion
      <a href={file} target={'_blank'} rel="noopener noreferrer">
```

packages/components/src/renderer/FileMessage/index.tsx
```diff
@@ -1,53 +1,52 @@
...
+  return (
+    <div>
+      <span>{title}</span>
+      <img src={image} alt={title} />
+    </div>
+  )
+}
+
+const useMimeType = (url: string): string | undefined => {
+  return useMemo(() => mimeTypes.getType(path.extname(url)) || undefined, [url])
```
laurentlp, line 51: We added this section to the code that fetches the Mime Type because otherwise, query parameters in URLs would give us undefined results. See this PR for more details: https://github.com/botpress/botpress/pull/5401

```ts
const validUrl = new URL(url)
extension = validUrl.pathname
```

packages/components/src/typings.ts
```diff
@@ -0,0 +1,57 @@
...
+  shouldPlay?: boolean // used for voice message only
+  onSendData: (data: any) => Promise<void>
+  onFileUpload: FileUploadHandler
+  onAudioEnded?: (this: HTMLMediaElement, ev: HTMLMediaElementEventMap['ended']) => void
+}
+
+export interface StudioConnector {
+  /** Event emitter */
+  events: any
+  /** An axios instance */
```
laurentlp, line 44: You can remove this comment IMHO.

packages/components/tsconfig.json
```diff
@@ -16,8 +16,12 @@
     "forceConsistentCasingInFileNames": true,
     "isolatedModules": true,
     "noImplicitAny": false,
-    "baseUrl": "./src"
+    "declaration": true,
+    "baseUrl": "./src",
+    "paths": {
+      "@botpress/messaging-server/content-types": ["../../server/src/content/types.ts"]
```
laurentlp, line 22: Is this still necessary?

packages/components/src/ReactRenderer.stories.tsx
```diff
@@ -0,0 +1,79 @@
...
+      control: {
+        type: 'object'
+      }
+    }
+  }
+} as ComponentMeta<typeof ReactMessageRenderer>
+
+const Template: ComponentStory<typeof ReactMessageRenderer> = (args) => <ReactMessageRenderer {...args} />
+
+export const Primary = Template.bind({})
```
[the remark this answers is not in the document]
charlescatta: You gotta click set object and write a JSON payload in the field that will appear. I'll change it to have a default value. This shows that the component throws if the input payload is invalid

packages/components/src/renderer/FileMessage/index.tsx
charlescatta: ok I fixed it, if a mime type isn't detected I'm reverting to a File content type. Is this good behavior or should I just try to render the content with a default mime type?

packages/components/tsconfig.json
charlescatta: nope, removed

packages/components/src/renderer/FileMessage/index.tsx
```diff
@@ -47,6 +53,13 @@ export const Image: React.FC<MessageTypeHandlerProps<'image'>> = ({ image, title
...
-const useMimeType = (url: string): string | undefined => {
-  return useMemo(() => mimeTypes.getType(path.extname(url)) || undefined, [url])
+const useMimeType = (url: string): string | null => {
+  return useMemo(() => {
+    try {
+      const validUrl = new URL(url)
+      return mimeTypes.getType(validUrl.pathname)
+    } catch (error) {
+      return null
+    }
```
laurentlp, line 63:
```suggestion
    let extension = ''
    try {
      const validUrl = new URL(url)
      extension = validUrl.pathname
    } catch (error) {
      extension = path.extname(url)
    }
    
    return mimeTypes.getType(extension)
```

My bad, I forgot that we had to push a fix for the PR I linked (fix: https://github.com/botpress/botpress/pull/5488). 

The result is a blend of using `new URL` to try to decode the URL and then find the file's extension. In case the URL is not valid (e.g. missing an http://) but still contains an extension, we then use path.extname() to extract it.

packages/components/src/renderer/FileMessage/index.tsx
laurentlp: See my last comment. Sorry for misleading you!

packages/components/src/content-typings.ts
charlescatta: The reason I did it like this is to be able to use the array of content types at runtime, this behaviour is no longer required (except for one test) but I don't see a reason why we wouldn't keep it?

packages/components/src/renderer/Dropdown/index.tsx
charlescatta: fixed

packages/components/src/renderer/FileMessage/index.tsx
charlescatta: fixed

packages/components/src/content-typings.ts
laurentlp: Fair enough

laurentlp (approved):
LGTM! Nice work!!!!

samuelmasse (approved):
Good job. Let's merge this rigth now as I'm currently working on a branch that contains two different PRs in it and it's a bit of a mess. This can go into master rigth away so I'll be able to push this into the webchat branches

PR #240, opened 2021-11-11
fix(server): fix response to other node

samuelmasse:
This PR fixes a situation in converse or websocket where sending a response message to a node other than the node where the converse request was made or where the websocket is connected would result in the response being lost.

Closes DEV-1977


packages/server/src/converse/service.ts
```diff
@@ -33,6 +40,37 @@ export class ConverseService extends Service {
     }
 
     const incomingId = this.incomingIdCache.get(message.id)
+    await this.distributed.publish(`converse/${incomingId}`, {
+      cmd: ConverseCmds.Message,
+      data: { message, incomingId }
```
laurentlp, line 45: Do we really need to send the incomingId considering that the subscriber already knows it?

packages/server/src/distributed/base/subservice.ts
```diff
@@ -4,6 +4,8 @@ export interface DistributedSubservice {
   setup(): Promise<void>
   destroy(): Promise<void>
   listen(channel: string, callback: (message: any) => Promise<void>): Promise<void>
```
laurentlp, line 6: Why not use subscribe here? It would make more sense considering that there is a function called unsubscribe that removes the listener.

packages/server/src/distributed/local/subservice.ts
```diff
@@ -15,6 +15,10 @@ export class LocalSubservice implements DistributedSubservice {
 
   async listen(channel: string, callback: (message: any) => Promise<void>) {}
 
+  async unsubscribe(channel: string) {}
+
+  async publish(channel: string, message: any) {}
```
laurentlp, line 20: It would be nice to have typings for the message here. Something generic like this that allows the caller to defined its own typings:

```suggestion
  async publish<Message = any>(channel: string, message: Message) {}
```

packages/server/src/distributed/service.ts
```diff
@@ -30,6 +30,14 @@ export class DistributedService extends Service {
     return this.subservice.listen(channel, callback)
   }
 
+  async unsubscribe(channel: string) {
+    return this.subservice.unsubscribe(channel)
+  }
+
+  async publish(channel: string, message: any) {
```
laurentlp, line 37: Same here as my comment above

packages/server/src/stream/service.ts
```diff
@@ -135,4 +145,14 @@ export class StreamService extends Service {
...
   }
+
+  private async getChannel(conversationId: uuid) {
+    const convmaps = await this.mapping.convmap.listByConversationId(conversationId)
+    if (convmaps.length === 1) {
+      const tunnel = await this.mapping.tunnels.get(convmaps[0].tunnelId)
+      return this.channels.getById(tunnel!.channelId).name
+    } else {
+      return 'messaging'
+    }
```
laurentlp, line 156: Here it might be nice to differentiate a multi-channel conversation from a conversation on an unknown channel. What do you think?

laurentlp (commented):
Nice work as usual. I think there is some refactor needed, but it could be done in another PR.

packages/server/src/converse/service.ts
samuelmasse: This is builtin in dispatcher

packages/server/src/distributed/base/subservice.ts
samuelmasse: Good point changed it to subscribe

packages/server/src/distributed/local/subservice.ts
samuelmasse: We have typings now for the Dispatcher. I don't see much use in putting typings in the DistributedService, as this service doesn't know what it has to send. As far as it's concerned it's sending some payload.

packages/server/src/stream/service.ts
samuelmasse: Maybe but this isn't a change done in this PR so it's not necessary to change this here. But let's keep this in mind.

laurentlp (approved):

packages/server/src/stream/dispatch.ts
```diff
@@ -0,0 +1,17 @@
...
+export enum StreamDispatches {
+  Message
+}
+
+export interface StreamMessageDispatch {
+  source?: string
+  data: {
+    type: string
+    data: any
+  }
```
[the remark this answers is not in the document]
samuelmasse: Agreed I renamed it to payload

PR #241, opened 2021-11-12
chore(server): add a benchmark script

samuelmasse:
Adds a script to benchmark the message creation route.

### DIFF
--- a/benchmark.sh
+++ b/benchmark.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+
+# To run this :
+# yarn global add loadtest
+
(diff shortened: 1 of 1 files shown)


laurentlp:
@samuelmasse What is the need behind this script?

samuelmasse:
@laurentlp I previously ran this benchmark to get a rough idea of how fast messaging is on my pc. Then I made the fix #240 that adds two pub call to redis for each message that gets sent and I wanted to test if this had a performance impact. Turns out it doesn't redis is super fast.

laurentlp (approved):

PR #245, opened 2021-11-12
chore(webchat): use components

samuelmasse:
Replaces the webchat's renderers with the renderers from the components package.

Refresh of #244 because something terribly wrong happened with the commit history on that one

Closes DEV-1884


packages/components/src/renderer/index.tsx
```diff
@@ -39,7 +39,8 @@ export class Renderer {
   private handlers: Partial<Record<MessageType, MessageTypeHandler<MessageType>>> = {}
 
   constructor() {
-    this.set('unsupported', ({ type }) => <div>Unsupported message type: {type}</div>)
+    // TODO: is this "type" property actually there? It's not in the typings.
+    this.set('unsupported', ({ type }: any) => <div>Unsupported message type: {type}</div>)
   }
```
samuelmasse, line 44: See comment

packages/components/src/renderer/index.tsx
```diff
@@ -48,7 +49,11 @@ export class Renderer {
 
   public register(handlers: Partial<{ [key in MessageType]: MessageTypeHandler<key> }>) {
     for (const type in handlers) {
-      this.set(type as MessageType, handlers[type])
+      this.set(
+        type as MessageType,
+        // TODO: fix typings here. This produces an implicit any
+        (handlers as any)[type]
```
samuelmasse, line 55: There's voodoo typings here that I couldn't make work in 5 minutes

packages/components/src/renderer/index.tsx
```diff
@@ -79,12 +84,15 @@ const defaultRenderer = new Renderer()
...
+  'custom',
+  ({ module, component, config, payload }: { module: any; component: any; config: any; payload?: any }) => {
+    if (module === 'extensions' && component === 'Dropdown') {
+      return <Dropdown {...payload} config={config} />
+    }
+    return <Custom module={module} component={component} config={config} />
   }
-  return <Custom module={module} component={component} config={config} />
-})
+)
```
samuelmasse, line 95: I guess this is supposed to be removed as stated by the comment. It would still be nice however to have typings. I had to put any there because of the noImplicitAny setting

packages/components/tsconfig.json
```diff
@@ -1,24 +1,26 @@
 {
+  // Manage references here
+  "references": [],
+
+  // Defaults (don't change this)
+  "extends": "../../tsconfig.packages.json",
   "compilerOptions": {
-    "experimentalDecorators": true,
-    "esModuleInterop": true,
+    "rootDir": "src",
```
samuelmasse, line 8: Notice the standard tsconfig file for the repo.

packages/webchat/src/components/messages/Message.tsx
```diff
@@ -1,230 +0,0 @@
-import classnames from 'classnames'
-import pick from 'lodash/pick'
-import { inject, observer } from 'mobx-react'
-import React, { Component } from 'react'
```
samuelmasse, line 4: I suppose we can remove this entire file?

packages/webchat/src/components/messages/MessageGroup.tsx
```diff
@@ -91,47 +54,39 @@ class MessageGroup extends React.Component<Props> {
...
+                    onFileUpload: this.props.onFileUpload,
+                    store: this.props.store,
+                    onAudioEnded: this.onAudioEnded,
+                    shouldPlay: this.state.audioPlayingIndex === i,
+                    intl: this.props.store!.intl,
+                    escapeHTML: true,
+                    isInEmulator: false,
+                    showTimestamp: this.props.store!.config.showTimestamp
+                  }}
                 />
```
samuelmasse, line 90: Here's the integration in the webchat. There doesn't seem to be an InlineFeedback option was this moved somewhere else? Also some props to have direct matches. For example where does `fromLabel` go?

packages/components/package.json
```diff
@@ -3,14 +3,14 @@
...
+  "source": "src/index.tsx",
   "license": "AGPL-3.0",
   "dependencies": {
     "html-truncate": "^1.2.2",
     "mime": "^2.5.2",
     "prop-types": "^15.7.2",
-    "react": "^16.8.6",
-    "react-dom": "^16.8.6",
+    "react": "^17.0.2",
+    "react-dom": "^17.0.2",
```
samuelmasse, line 13: I changed the react version back to 17 because typings were messing up somewhere.

packages/components/package.json
```diff
@@ -19,8 +19,8 @@
     "snarkdown": "^2.0.0"
   },
   "scripts": {
-    "build": "rollup -c",
-    "watch": "rollup --watch -c",
+    "build": "yarn --silent && tsc --build",
+    "watch": "yarn --silent && tsc --build --watch",
```
samuelmasse, line 23: rollup was removed. Here the scripts are purely to test the typings with tsc. If we want to build a bundle we can use `parcel build`, but considering the components package is a npm module I don't think we need to bundle anything. Simply distributing what tsc produces should work fine.

packages/components/tsconfig.json
```diff
@@ -1,24 +1,26 @@
...
-    "skipLibCheck": false,
-    "allowSyntheticDefaultImports": true,
-    "strict": true,
-    "forceConsistentCasingInFileNames": true,
-    "isolatedModules": true,
-    "noImplicitAny": false,
-    "declaration": true,
+
+    // TODO: remove this
     "baseUrl": "./src"
```
samuelmasse, line 22: We don't use baseUrl in this repo. I put it there for now to keep the PR small. I'll make a different PR that removes this. Basically the reasoning behind this is that having a baseUrl (or any sort of paths config) creates multiple different ways of importing a source file. Some ways are preffered over others, and the IDE doesn't always suggest the best one, so we either waste time in PRs asking to import the file in the correct way, or we don't bother with that and the codebase becomes a mess of different preferences of how that or this programmer prefers to imports their files (see https://github.com/botpress/botpress).

You can see as well that the codebase doesn't make any use of `export default` and doesn't use any `index` files except for one index file at the root of every package to control the exports. The same thing could be done in the components package. There isn't any good reason in my opinion that files that imports a class should decide to name it to whatever it wants. Instead of going with the "ok I'll name the export from this file like this everywhere I import it" strategy we can just leverage the typescript compiler to check for us that we're importing the correct thing by using named exports.

Let me know what you think

samuelmasse:
@charlescatta I started the integration of the components package in the webchat port. Rigth now there seems to be css issues as I'm not getting any styling.

I removed the dependency on rollup, as we'll now use parcel to develop the webchat. I completely changed the tsconfig to be base on the template we have for the repo. I also added a noImplicitAny constraint in the process that caused 4-5 issues in the code (that I partially solved).

packages/components/src/renderer/index.tsx
charlescatta: The payload of a custom component can't be known at build time

packages/components/src/renderer/index.tsx
charlescatta: fixed

packages/components/src/renderer/index.tsx
charlescatta: fixed

packages/components/src/renderer/Text/index.tsx
```diff
@@ -1,4 +1,5 @@
 import React from 'react'
+// @ts-ignore
```
laurentlp, line 2: Reason behind this?

packages/components/tsconfig.json
laurentlp: Solid argument, I agree, let's go that way!

packages/webchat/src/components/OverridableComponent.tsx
```diff
@@ -32,7 +32,7 @@ class OverridableComponent extends React.Component<Props, State> {
     return this.props.overrides?.[this.props.name]
       ?.map(({ module, component }) => ({
         key: `${module}:${component}`,
-        element: window.botpress[module]?.[component]
+        element: (window.botpress as any)[module]?.[component]
```
laurentlp, line 35: Why do you need to cast this as any?

packages/webchat/src/components/messages/Message.tsx
```diff
@@ -173,11 +41,33 @@ class Message extends Component<MessageProps> {
...
+          isBotMessage: this.props.isBotMessage!,
+          sentOn: this.props.sentOn!,
+          onSendData: this.props.onSendData!,
+          onFileUpload: this.props.onFileUpload!,
+          store: this.props.store,
+          onAudioEnded: this.props.onAudioEnded,
+          shouldPlay: this.props.shouldPlay,
+          intl: this.props.store!.intl,
+          escapeHTML: true,
+          isInEmulator: false,
```
laurentlp, line 65: I don't think this should be a constant! What will happen once we use this version of the webchat as the emulator?

packages/webchat/src/index.tsx
```diff
@@ -57,6 +57,9 @@ export { Embedded as embedded } from '.'
 export { Fullscreen as fullscreen } from '.'
 
 export * from './typings'
+
+// TODO: export this
+/*
```
laurentlp, line 62: Why not simply remove the comments then?

packages/webchat/src/typings.ts
```diff
@@ -19,9 +19,6 @@ declare global {
     USE_SESSION_STORAGE: boolean
     // TODO: this should be of type BPStorage
     BP_STORAGE: any
-    botpress: {
-      [moduleName: string]: any
-    }
```
laurentlp, line 24: Reason behind this change?

packages/webchat/src/components/messages/MessageGroup.tsx
```diff
@@ -91,16 +89,12 @@ class MessageGroup extends React.Component<Props> {
...
-              /*
-              // TODO: use the components package
-              if (payload?.wrapped) {
-                payload = { ...payload, wrapped: renderPayload(payload.wrapped) }
-              } else {
-                payload = renderPayload(payload)
+              let payload = this.convertPayloadFromOldFormat(message)
+
+              // TODO: remove this. quick_reply should be single-choice
+              if (payload.type === 'single-choice') {
```
laurentlp, line 95: Any idea what is the root cause of this?

laurentlp (commented):
Small comments but overall LGTM!

packages/components/src/renderer/Text/index.tsx
samuelmasse: react-text-format doesn't have typings

packages/webchat/src/components/OverridableComponent.tsx
samuelmasse: Otherwise it creates an implicit any. So we explicitly cast to any instead.

packages/webchat/src/components/messages/Message.tsx
samuelmasse: Rigth there's an options passed in the props for that. Changed it

packages/webchat/src/typings.ts
samuelmasse: There's a declaration of botpress already in the components package

packages/webchat/src/components/messages/MessageGroup.tsx
samuelmasse: We send a message of type 'single-choice' and when the user clicks on a button it returns a message of type quick_reply. So the components package simply uses the wrong name. I'll fix this problem in another PR

packages/webchat/src/index.tsx
samuelmasse: This is a reminder for the future, we'll keep it here for now but it'll be removed/uncommented eventually.

laurentlp: Alright, then what's the goal of having those exported?

laurentlp (approved):

PR #246, opened 2021-11-15
chore(components) no base url

samuelmasse:
Removes the baseUrl options in the tsconfig for the components package. As expected the code wasn't consistent in how it imports the files (using the base url or not) so removing this option will clean things up

Closes DEV-1884


laurentlp (approved):
LGTM!

PR #247, opened 2021-11-16
chore(components): no default exports no index files

samuelmasse:
Removes usage of default exports and index files where possible in the components package.

Closes DEV-1884


laurentlp (approved):
LGTM

PR #249, opened 2021-11-18
chore(engine): create engine package

samuelmasse:
Creates an engine package.

Closes DEV-1977


samuelmasse:
Docker is failing not sure why. It seems that it wants the engine package to be hosted on npm to build the server. Preferably we don't host this on npm so we should figure out a way to not have to do that.

laurentlp (approved):

PR #254, opened 2021-11-23
feat(socket): authenticate in handshake

samuelmasse:
Requires that authentication be provided when a user connects to a socket. A socket connection can no longer be held without first authenticating a user.

We'll need to bump the socket package version before merging this

Closes DEV-2074

### DIFF
--- a/packages/server/src/socket/manager.ts
+++ b/packages/server/src/socket/manager.ts
@@ -106,6 +106,8 @@ export class SocketManager {
             (await this.userTokens.verifyToken(userTokenId, userTokenToken))
           ) {
             socket.data.creds = creds
+            // we don't need to send it back if it was already sent to us
+            delete socket.data.creds.userToken
             return next()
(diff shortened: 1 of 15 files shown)


packages/server/src/socket/manager.ts
```diff
@@ -76,12 +79,62 @@ export class SocketManager {
...
+      if (creds) {
+        const user = await this.users.get(creds.userId)
+        if (user?.clientId === clientId) {
+          const [userTokenId, userTokenToken] = creds.userToken.split('.')
+          if (
+            validateUuid(userTokenId) &&
+            userTokenToken?.length &&
+            (await this.userTokens.verifyToken(userTokenId, userTokenToken))
+          ) {
+            socket.data.creds = creds
```
laurentlp, line 108: Remove token from creds

laurentlp (approved):

PR #255, opened 2021-11-23
fix(mapping): fix insert race conditions

samuelmasse:
It's sometime possible that sometimes the server checks if a mapping exists, and then creates it if it doesn't. If this is done fast enough, it's possible that the check to see if the mapping exists returns false more than once, and the process of creating the mapping starts multiple times. The constraints in the db will prevent more than once request from succeeding, meaning that exceptions will be thrown.

This PR fixes this problem

Closes DEV-1618

### DIFF
--- a/packages/server/src/mapping/convmap/service.ts
+++ b/packages/server/src/mapping/convmap/service.ts
@@ -19,6 +19,7 @@ export class ConvmapService extends Service {
   private table: ConvmapTable
   private cacheByThreadId!: ServerCache2D<Convmap>
   private cacheByConversationId!: ServerCache<uuid, Convmap[]>
+  private locks!: ServerCache2D<Promise<Convmap>>
 
   constructor(
     private db: DatabaseService,
@@ -35,6 +36,7 @@ export class ConvmapService extends Service {
   async setup() {
     this.cacheByThreadId = await this.caching.newServerCache2D('cache_convmap_by_thread_id')
     this.cacheByConversationId = await this.caching.newServerCache('cache_convmap_by_conversation_id')
+    this.locks = await this.caching.newServerCache2D('cache_convmap_locks')
 
     this.batcher = await this.batching.newBatcher(
       'batcher_convmap',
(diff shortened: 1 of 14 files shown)


packages/server/test/tsconfig.json
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "../../../tsconfig.packages.json",
+  "compilerOptions": {
+    "rootDir": ".."
+  },
+  "include": ["../test/**/*", "../src/**/*"]
+}
```
samuelmasse, line 7: @laurentlp did you remove this config? It's necessary to have intelisense with vscode

laurentlp: Right, I'll make sure to keep it!

packages/server/test/mapping.test.ts
```diff
@@ -0,0 +1,90 @@
+import { uuid } from '@botpress/messaging-base'
```
laurentlp, line 1: PP: IMHO it's better when tests take the same structure (or so) as the `src` folder

laurentlp (commented):
Just make sure to comment on the reason behind this semi-hack of having a cache of promise so we don't forget in the future. 

Also, we talked about adding a retry in the mapping service. Is this out of the equation?

packages/server/src/mapping/convmap/service.ts
```diff
@@ -97,6 +104,19 @@ export class ConvmapService extends Service {
...
+      return convmap
+    }
+
+    return this.barrier.once(tunnelId, threadId, async () => {
+      const tunnel = await this.tunnels.get(tunnelId)
+      const conversation = await this.conversations.create(tunnel!.clientId, userId)
+      return this.create(tunnelId, conversation.id, threadId)
+    })
+  }
+
```
samuelmasse, line 119: @laurentlp here's a barrier in use. The code in the callback will be excuted only once, and return the same value for all Promises waiting on it. The way a single "task" is identified is with arguments before the callback. This particular barrier is a 2d barrier (similarly to our 2d caches), so the key is composed of two keys.

laurentlp (approved):
LGTM! Nice work :ok_hand:

PR #256, opened 2021-11-24
chore(test): added unit, integration and e2e tests config and setup

laurentlp:
This PR separates all types of tests in different categories (unit, integration, e2e) with each having its own configuration and setup.

Closes DEV-2047
Closes DEV-2048
Closes DEV-2052


test/jest.e2e.setup.ts
```diff
@@ -0,0 +1,24 @@
...
+require('ts-node/register')
+
+import { setup as setupDevServer } from 'jest-dev-server'
+import jestSetup from './jest.setup'
+
+const setup = async () => {
+  await jestSetup()
+
+  // TODO: Maybe put this somewhere else?
+  process.env.ENABLE_EXPERIMENTAL_SOCKETS = 'true'
```
samuelmasse, line 11: We'll find some way to avoid these global config in another PR.

samuelmasse (approved):
That's going to be a much better setup for our tests!

packages/server/package.json
```diff
@@ -8,7 +8,7 @@
     "postbuild": "node ./dist/build",
     "watch": "yarn --silent && tsc --build --watch",
     "start": "node dist/index.js",
-    "dev": "yarn --silent && ts-node-dev --inspect --inspect-publish-uid=http --transpile-only src/index.ts"
+    "dev": "yarn --silent && ts-node-dev --debounce 500 --transpile-only src/index.ts"
```
samuelmasse, line 11: We were starting an inspector on the same port every time which prevented two instances of the server to run on yarn dev at the same time. I initially did this so we could attach the debugger all the time, but I ended up never using that so I removed it here. The debouce of 500ms prevents the server from restarting multiple times if you do CTRL + S a lot

PR #258, opened 2021-11-25
fix(redis): fix master failure crashing the server

samuelmasse:
There was a problem when using redis with sentinel, master and slaves where switching a master could make the server crash. The server was set to throw an exception is an attempt to connect failed more than 10 times. I removed this limit so now the server can retry to connect to redis indefinitely.

Closes DEV-2002


laurentlp:
@samuelmasse Can you please give me the steps required to reproduce the issue so I can test it on my end!

laurentlp (approved):

PR #260, opened 2021-11-26
fix(components): fix carousel component

laurentlp:
This PR fixes the carousel component by handling empty carousel actions. It also exports the default config and some interfaces so it can be used by anyone. 

Closes DEV-2045


samuelmasse (approved):
yes please

PR #261, opened 2021-11-26
fix(monitoring): fix listing outdated errored conduits

samuelmasse:
Prevents conduits that have reached the maximum amount of errors from being fetched again when looking to initialize outdated conduits

Closes DEV-1631


packages/server/src/status/service.ts
```diff
@@ -16,70 +17,102 @@ export class StatusService extends Service {
...
+          // we exclude lastError because it migth make the query slow
+          .select('conduitId', 'numberOfErrors', 'initializedOn')
+          .where('numberOfErrors', '<', maxAllowedFailures)
+          .andWhere((q) =>
+            q
+              .where('initializedOn', '<', this.db.setDate(new Date(Date.now() - tolerance))!)
+              .orWhereNull('initializedOn')
+          )
+          .limit(limit)
+      ).map((x) => this.deserialize(x))
```
laurentlp, line 114: What if the query returns no result?

laurentlp (commented):
We will need to have a migration for this PR but otherwise, it looks fine to me. Nice work!

packages/server/src/status/service.ts
samuelmasse: It that case the query returns an empty array, so the .map still works

samuelmasse:
@laurentlp The migration is likely going to be just dropping the table. The status table doesn't store anything permanent so it's not really required to bother porting everything.

laurentlp:
> @laurentlp The migration is likely going to be just dropping the table. The status table doesn't store anything permanent so it's not really required to bother porting everything.

You also edited the conduit table. I think it would be 1000 times better if we would do this the proper way. It can be done in another PR, but I strongly suggest we do both migrations.

packages/server/src/migrations/0.1.18-status.ts
```diff
@@ -0,0 +1,35 @@
...
+  }
+
+  async down() {
+    await this.trx.schema.dropTableIfExists('msg_status')
+
+    return this.trx.schema.alterTable('msg_conduits', (table) => {
+      table.timestamp('initialized').nullable()
+      table.index(['initialized'])
+    })
+  }
```
samuelmasse, line 34: @laurentlp I think this migration is good enough. It performs the changes on the msg_conduit table and just deletes the status table. The status table contains volatile data anyways so why bother doing all the changes

packages/server/src/migrations/0.1.18-status.ts
```diff
@@ -0,0 +1,35 @@
...
+
+  async down() {
+    await this.trx.schema.dropTableIfExists('msg_status')
+
+    return this.trx.schema.alterTable('msg_conduits', (table) => {
+      table.timestamp('initialized').nullable()
+      table.index(['initialized'])
+    })
```
laurentlp, line 33: I understand that `applied` protects the up migration in case it was already executed but there is no similar mechanism for the down migration. 

I think we should maybe pass an argument to applied (up/down) and execute different queries depending on the context. What do you think? The same could apply to valid!

packages/server/test/functional/status.test.ts
```diff
@@ -0,0 +1,220 @@
...
+
+    expect(outdateds).toEqual([state.status, state.status2])
+  })
+
+  test('Add error', async () => {
+    await status.addError(state.conduit.id, new Error('err1'))
+
+    const st = await status.get(state.conduit.id)
+    expect(st!.numberOfErrors).toBe(1)
+    expect(st!.lastError?.startsWith('Error: err1')).toBeTruthy()
```
laurentlp, line 103:
```suggestion
    const errMessage = 'err1'
    await status.addError(state.conduit.id, new Error(errMessage))

    const st = await status.get(state.conduit.id)
    expect(st!.numberOfErrors).toBe(1)
    expect(st!.lastError?.contains(errMessage)).toBeTruthy()
```

laurentlp: Same applies to the tests below.

packages/server/test/functional/status.test.ts
```diff
@@ -0,0 +1,220 @@
...
+
+    state.status = st
+  })
+
+  test('List outdated should not contain errored conduit', async () => {
+    // conduit currently has 5 registered errors so it should no be listed
+    const maxAllowedErrors = 5
+    const outdateds = await status.listOutdated(ms('1h'), maxAllowedErrors, 10)
+
+    expect(outdateds).toEqual([state.status2])
```
laurentlp, line 135: I think it would be best if you were to check that outdated does not contain errored conduits!

packages/server/test/functional/status.test.ts
```diff
@@ -0,0 +1,220 @@
...
+  })
+
+  test('List outdated should contain both conduits if max allowed errors is higher', async () => {
+    const maxAllowedErrors = 6
+    const outdateds = await status.listOutdated(ms('1h'), maxAllowedErrors, 10)
```
laurentlp, line 140: Can you put the `10` in a global constant and same for the time value?

packages/server/test/functional/status.test.ts
```diff
@@ -0,0 +1,220 @@
...
+    const st = await status.get(state.conduit2.id)
+    expect(st!.initializedOn).toEqual(date)
+
+    state.status2 = st
+  })
+
+  test('List outdated should not contain both initialized conduits', async () => {
+    const outdateds = await status.listOutdated(ms('1h'), 5, 10)
+
+    expect(outdateds).toEqual([])
```
laurentlp, line 197: PP:
```suggestion
    expect(outdateds.length).toEqual(0)
```

packages/server/test/functional/status.test.ts
```diff
@@ -0,0 +1,220 @@
...
+
+    state.status = st
+  })
+
+  test('List outdated should include the uninitialized conduit again', async () => {
+    const outdateds = await status.listOutdated(ms('1h'), 5, 10)
+
+    expect(outdateds).toEqual([state.status])
+  })
+})
```
laurentlp, line 220: Add a test that makes sure we only return X status where X is the batch size

packages/server/src/migrations/0.1.18-status.ts
samuelmasse: applied is use for both up and down. This is the code for the base migration class. We just inverse the condition depending if it's up or down :

```ts
  async shouldRun() {
    if (!(await this.valid())) {
      return false
    }

    if (this.isDown) {
      return this.applied()
    } else {
      return !(await this.applied())
    }
  }
```

packages/server/src/migrations/0.1.18-status.ts
```diff
@@ -0,0 +1,35 @@
...
+    name: StatusMigration.name,
+    description: 'Moves status information from conduit table to the status table',
+    version: '0.1.19'
+  }
+
+  async valid() {
+    return this.trx.schema.hasTable('msg_conduits')
+  }
+
+  async applied() {
```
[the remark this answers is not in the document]
samuelmasse: No see my other comment

packages/server/test/functional/status.test.ts
samuelmasse: This check does that and also checks that other stuff returned is correct so it's harder to have a false positive here

samuelmasse:
Note : remove version bump before merging

laurentlp (approved):
LGTM!!! :100:

PR #266, opened 2021-12-01
feat(channels): add conversation.started event

samuelmasse:
Adds a conversation.started event that allows sending a proactive message

Closes DEV-1623
Closes DEV-2070

### DIFF
--- a/packages/server/src/channels/base/conduit.ts
+++ b/packages/server/src/channels/base/conduit.ts
@@ -80,6 +80,7 @@ export abstract class ConduitInstance<TConfig, TContext extends ChannelContext<a
 
   async receive<T>(payload: T, started?: boolean) {
     const conduit = (await this.app.conduits.get(this.conduitId))!
+    const channel = this.app.channels.getById(conduit!.channelId)
     const provider = (await this.app.providers.getById(conduit.providerId))!
     const endpoint = await this.extractEndpoint(payload)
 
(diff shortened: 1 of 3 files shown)


packages/server/src/channels/base/conduit.ts
```diff
@@ -96,9 +97,22 @@ export abstract class ConduitInstance<TConfig, TContext extends ChannelContext<a
...
+        userId,
+        {
+          conduit: { id: this.conduitId, endpoint }
+        }
+      )
+    } else {
+      return this.app.messages.create(conversationId, userId, endpoint.content, {
+        conduit: { id: this.conduitId, endpoint: _.omit(endpoint, 'content') }
+      })
+    }
```
samuelmasse, line 115: Not the cleanest implementation. Channels are going to be refactored when they are moved to their own package and this is going to be way better

samuelmasse:
@laurentlp Should we consider the /start command to be a proactive message? This might break users. We could make this change in the v2 version of the channel

Screenshot from 2021-12-01 14-36-53

packages/server/src/channels/teams/conduit.ts
```diff
@@ -72,6 +72,8 @@ export class TeamsConduit extends ConduitInstance<TeamsConfig, TeamsContext> {
         await turnContext.sendActivity(message)
       })
     }
+
+    await this.receive(turnContext, true)
```
laurentlp, line 76: We send the event after sending the proactive message to teams?

laurentlp:
> @laurentlp Should we consider the /start command to be a proactive message? This might break users. We could make this change in the v2 version of the channel
> 
> Here the bot would not respond for example and you would need to send another message. The thing is that the /start command is sent automatically when you start a conversation, so it is sort of a proactive trigger

I think we have to in order to support proactive messages on Telegram. Since we don't have access to any other hook telling us that a user created a new conversation with a given bot, we should use the `/start` command as the `conversation.started` event.

Also, is my understanding of this feature correct? We would receive a `/start` message, then send a `conversation.started` event which would allow the user to send a proactive message or discard it and go with the normal flow?

samuelmasse:
@laurentlp Rigth now we treat /start as a text message containing `/start`. Which is why I suggest we move the /start command to be a proactive trigger in the next version of the channel otherwise we migth break many people's implementation.

packages/server/src/channels/teams/conduit.ts
samuelmasse: The proactive message implementation above is a legacy implementation. The new way of doing this is receiving the proactive trigger in botpress and sending a message. We keep the old one as well to avoid breaking people's stuff.

laurentlp (approved):

PR #267, opened 2021-12-01
chore(components): improve components dependency management and minor changes

laurentlp:
This PR fixes some issues with the components package that we caused by `react-intl@2`. Since this version of the package required to install an external lib for its typing, it could create conflicts with a more recent version of the package that now comes with its own typings. 

I also added a `.npmignore` file to make sure we only publish relevant files to NPM.


packages/components/.npmignore
```diff
@@ -0,0 +1 @@
+.tsbuildinfo
```
samuelmasse, line 1: .npmignore? There isn't a yarn equivalent?

packages/components/package.json
```diff
@@ -41,13 +38,29 @@
...
     "@types/html-truncate": "^1.2.1",
     "@types/mime": "^2.0.3",
     "@types/react": "^17.0.34",
     "@types/react-dom": "^17.0.11",
-    "@types/react-intl": "^2.3.18",
     "@types/react-slick": "^0.23.5",
     "jest-css-modules-transform": "^4.3.0",
+    "prop-types": "^15.7.2",
+    "react": "^17.0.2",
+    "react-dom": "^17.0.2",
```
samuelmasse, line 50: React is a dev dependency?

packages/components/package.json
```diff
@@ -28,6 +15,16 @@
...
   ],
+  "dependencies": {
+    "html-truncate": "^1.2.2",
+    "mime": "^2.5.2",
+    "react-intl": "^3.12.1",
+    "react-linkify": "^0.2.2",
+    "react-select": "^5.1.0",
+    "react-slick": "^0.24.0",
+    "react-text-format": "^2.0.28",
+    "snarkdown": "^2.0.0"
```
samuelmasse, line 26: So these are both dependencies and peer dependencies?

samuelmasse (approved):
I'm a bit confused about the dependencies here. Why are some of them in more than one categore (dependencies, devDependencies, peerDependencies)?

packages/components/.npmignore
laurentlp: Yarn will also read this file, and I don't think a `.yarnignore` file exists.

packages/components/package.json
laurentlp: Having `react` as a dev dependency and a peer dependency allows for any project using this library to not necessarily rely on v17.0.2 but rather use "any compatible" version they are already using.

For example, you might want to include this component library into your react app that runs v17.2.0. In this case, it would make no sense to also install 17.0.2 since the lib is compatible with any version above 17 (as per de peer dependency).

packages/components/package.json
laurentlp: Exactly, my end goal here is to make sure that we have access to those libraries with versions that are compatible with our code. But at the same time, if you have, let's say, `snarkdown@2.9.9` installed and it's compatible, we'll use this version instead of downloading the one specified in the dependency section.

Peer dependencies is a way of saying we support all those versions for this particular library. If you also have a supported version installed, we'll use it otherwise, install this version that comes from our dependency list.

packages/components/src/typings.ts
```diff
@@ -1,13 +1,8 @@
+/// <reference lib="dom" />
```
laurentlp, line 1: This is the magic that tells Typescript to include the dom library if it ever needs to interpret this file.

PR #272, opened 2021-12-03
chore(channels): use channels from channels package

samuelmasse:
Imports channels from the @botpress/messaging-channels packages and uses them as replacement for the old channel code that was inside the server code.

Closes DEV-2125


packages/server/package.json
```diff
@@ -27,6 +27,7 @@
   "dependencies": {
     "@botpress/messaging-base": "0.0.1",
     "@botpress/messaging-engine": "0.0.1",
+    "@botpress/messaging-channels": "0.0.1",
```
laurentlp, line 30: We should consider using '*' as the version since it references another package and we want to use it whatever version it is at

packages/server/src/instances/service.ts
```diff
@@ -80,59 +71,41 @@ export class InstanceService extends Service {
...
-
-    this.cache = await this.cachingService.newServerCache('cache_instance_by_conduit_id', {
-      dispose: async (k, v) => {
-        if (!this.destroyed) {
-          await this.handleCacheDispose(k, v)
-        }
-      },
-      max: 50000,
-      maxAge: ms('30min')
-    })
```
samuelmasse, line 93: We'll need to put back this logic of clearing channels after not being used for x amount of minutes. To be done in future PR.

packages/server/src/channels/api.ts
```diff
@@ -5,8 +5,39 @@ export class ChannelApi {
   constructor(private router: Router, private app: App) {}
 
   async setup() {
+    const whRouter = Router()
```
laurentlp, line 8: wh? What does it mean?

packages/server/src/channels/service.ts
```diff
@@ -1,23 +1,23 @@
...
 
 export class ChannelService extends Service {
   private table: ChannelTable
 
-  private channels: Channel<ConduitInstance<any, any>>[]
-  private channelsByName: { [name: string]: Channel<ConduitInstance<any, any>> }
-  private channelsById: { [id: string]: Channel<ConduitInstance<any, any>> }
+  private channels: Channel<any, any, any, any>[]
+  private channelsByName: { [name: string]: Channel<any, any, any, any> }
+  private channelsById: { [id: string]: Channel<any, any, any, any> }
```
laurentlp, line 20: That's a lot of any!

packages/server/src/instances/service.ts
```diff
@@ -141,44 +123,52 @@ export class InstanceService extends Service {
...
+    }
 
     try {
       await this.distributedService.using(`lock_dyn_instance_setup::${conduitId}`, async () => {
-        await instance.setup(conduitId, conduit.config, this.app)
+        await channel.start(provider.name, conduit.config)
       })
-      this.cache.set(conduitId, instance, channel.lazy && this.lazyLoadingEnabled ? undefined : Infinity)
-
       await this.emitter.emit(InstanceEvents.Setup, conduitId)
```
laurentlp, line 147: It would make more sense IMHO if we were to send events with the same name as the function that triggers them! So in this case, shouldn't we use Start instead of setup? Or maybe we should rename the function to setup instead of start? What do you think?

packages/server/src/instances/invalidator.ts
```diff
@@ -40,33 +36,33 @@ export class InstanceInvalidator {
...
   }
 
   private async onConduitDeleting(conduitId: uuid) {
-    this.cache.del(conduitId, true)
+    await this.instances.stop(conduitId)
   }
 
   private async onConduitUpdated(conduitId: uuid) {
-    this.cache.del(conduitId, true)
+    await this.instances.stop(conduitId)
```
samuelmasse, line 53: Notice here we were doing distributed invalidations on the cache. Now there is a dispatcher in the instance service that informs the other node when they need to stop their conduit

packages/server/src/sync/schema.ts
```diff
@@ -1,16 +1,16 @@
...
 import Joi from 'joi'
-import { Channel } from '../channels/base/channel'
 
 const SyncWebhookSchema = Joi.object({
   url: Joi.string().uri().required(),
   token: Joi.string()
 })
 
-export const makeSyncRequestSchema = (channels: Channel<any>[]) => {
+export const makeSyncRequestSchema = (channels: Channel<any, any, any, any>[]) => {
```
laurentlp, line 9: Again, too much any for me hahaha

packages/channels/src/telegram/api.ts
```diff
@@ -67,6 +76,11 @@ export class TelegramApi extends ChannelApi<TelegramService> {
     }
   }
 
+  private useWebhook() {
+    // TODO: remove this dependency on server env vars
+    return !yn(process.env.SPINNED) || yn(process.env.CLUSTER_ENABLED)
+  }
```
samuelmasse, line 82: In future PR

packages/server/src/channels/api.ts
samuelmasse: wh = webhook

I changed it to webhookRouter

packages/server/src/channels/service.ts
samuelmasse: Fixed this by adding a ChannelTemplate class that has all the generics and a Channel interface that just has parameters of type any

packages/server/src/sync/schema.ts
samuelmasse: Fixed

PR #274, opened 2021-12-03
fix(migrations): fix migrations with targets that are out of bounds

samuelmasse:
Fixes a bug where asking the server to migrate up to a version lower than the current server version, or migrate down to a version higher than the current server version, would result in 0 migrations getting run and the server version getting wrongly updated in the meta table

Closes DEV-2137


packages/server/src/migrations/index.ts
```diff
@@ -1,4 +1,4 @@
 import { Migration } from '@botpress/messaging-engine'
-import { StatusMigration } from './0.1.18-status'
+import { StatusMigration } from './0.1.19-status'
```
samuelmasse, line 2: This migration had the wrong file name, but it has the correct server version in it's metadata so it shouldn't have caused problems

laurentlp (commented):
So if you do an up migration to a version (v1.5.6) higher than the latest one, will it save v1.5.6 in the meta table? Same goes for the down migration.

Also, can you make sure to update the PR title so it better reflects what is this fix

samuelmasse:
@laurentlp Good point I added a check to prevent migration to a version higher than the version of the server's package.

samuelmasse:
For the down migration, you could technically migrate down to 0.0.0 but that wouldn't cause any missing migration so it's correct behavior

laurentlp (approved):

PR #275, opened 2021-12-06
chore(components): renamed quick_reply to single-choice and added quick_reply component

laurentlp:
This PR fixes a couple of issues.

First I renamed the current `quick_reply` component to `single-choice` as it displays single choices, not quick replies. I then added the quick reply component that basically is the equivalent of the text component.

I also added typings for the `react-text-format`.


samuelmasse (approved):
LGTM

PR #276, opened 2021-12-07
chore(ci): use botpress shared actions v1

laurentlp:
This PR pins the version of the action(s) we use from https://github.com/botpress/gh-actions to `v1`


samuelmasse (approved):
:+1:

PR #280, opened 2021-12-10
chore(repo): stream layer

samuelmasse:
Creates a seperate application layer called stream. It handles the webhook and socket callbacks. The PostService and StreamService were removed. The StreamService is now the stream layer, and the PostService was merged in the utility Streamer class

Closes DEV-2126


packages/server/src/base/streamer.ts
```diff
@@ -0,0 +1,127 @@
...
+        ...data
+      }
+    }
+
+    if (yn(process.env.LOGGING_ENABLED)) {
+      this.logger.debug(`${clc.blackBright(`[${clientId}]`)} ${clc.cyan(type)}`, payload)
+    }
+
+    if (userId) {
+      await this.dispatcher.publish(StreamerDispatches.Message, userId, { source: source?.socket?.id, payload })
```
laurentlp, line 68: Maybe we should check if the user is connected through socket(s) before firing this event? This would prevent us from "disturbing" all the other servers when it might not be necessary. What do you think?

laurentlp (approved):
Make sure to remove this test (`packages/server/test/unit/post.test.ts`) as it will be converted into a service test and is outdated with your new changes. Otherwise, I have two small comments but overall it looks good to me! Nice work!!!

packages/server/src/base/streamer.ts
samuelmasse: I think you missed a part of the code. The handleUserConnected and handleUserDisconnected subscribe and unsubscribe the node to a particular user's events. So we in fact only disturb nodes where the user in connected

packages/server/src/base/streamer.ts
```diff
@@ -0,0 +1,127 @@
...
+      type,
+      data: {
+        clientId,
+        ...(userId ? { userId } : {}),
+        ...data
+      }
+    }
+
+    if (yn(process.env.LOGGING_ENABLED)) {
+      this.logger.debug(`${clc.blackBright(`[${clientId}]`)} ${clc.cyan(type)}`, payload)
```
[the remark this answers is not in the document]
samuelmasse: 100%. To be done in logging refactor. I already have an implementation in mind, I might do this in next sprint

PR #283, opened 2021-12-10
chore(tests): improved http client e2e tests

laurentlp:
This PR adds more tests for the HTTP client.
 - It covers every public functions exposed by the client
 - It tests all possible routes (happy and unhappy paths)
 - It improves the structure of the test suite

Closes DEV-2054


packages/base/src/sync.ts
```diff
@@ -2,7 +2,7 @@ import { uuid } from './uuid'
 
 export interface SyncRequest {
   channels?: SyncChannels
-  webhooks?: SyncWebhook[]
+  webhooks?: Omit<SyncWebhook, 'token'>[]
```
laurentlp, line 5: We will have to publish a new version of the base package before merging this PR!

packages/client/src/messages.ts
```diff
@@ -44,7 +44,7 @@ export class MessageClient extends BaseClient {
     await this.http.post(`/messages/turn/${id}`)
   }
 
-  public deserialize(message: Message): Message {
+  private deserialize(message: Message): Message {
```
laurentlp, line 47: I put those functions private as they should not be exposed to the user IMHO.

packages/server/src/conversations/api.ts
```diff
@@ -58,6 +58,11 @@ export class ConversationApi {
   async recent(req: ClientApiRequest, res: Response) {
     const userId = req.params.userId as uuid
 
+    const user = await this.users.get(userId)
+    if (!user || user.clientId !== req.client.id) {
+      return res.sendStatus(404)
+    }
```
laurentlp, line 64: Otherwise, it would throw 500 HTTP errors since we might try to create a conversation with a userId that does not exist. This would cause the database to send foreign key constrain violations. I copied this piece of code from the list function above.

packages/server/src/conversations/service.ts
```diff
@@ -212,6 +213,7 @@ export class ConversationService extends Service {
       })
       .groupBy('msg_conversations.id', 'msg_messages.id')
       .orderBy('sentOn', 'desc')
+      .orderBy('createdOn', 'desc')
```
laurentlp, line 215: Since sentOn is null when listing conversations, I added a sort on createdOn.

packages/server/src/sync/schema.ts
```diff
@@ -2,8 +2,7 @@ import { Channel } from '@botpress/messaging-channels'
 import Joi from 'joi'
 
 const SyncWebhookSchema = Joi.object({
-  url: Joi.string().uri().required(),
-  token: Joi.string()
```
laurentlp, line 6: As discussed, we should not let the user send webhook tokens when syncing its configs.

samuelmasse (approved):
LGTM. I get a weird error when running the e2e tests though.

Screenshot from 2021-12-13 13-31-31

laurentlp:
> 

It will be fixed in another PR since it's only a warning

PR #284, opened 2021-12-13
fix(sync): add lock for provider name

samuelmasse:
Fixes a problem in sync calls when messaging is spinned where multiple calls made with the same botId would not be locked. Related https://github.com/botpress/botpress/pull/5760

Closes DEV-2189


samuelmasse:
We need to release 0.1.20 with that

laurentlp (approved):
Tested and working :+1:

PR #287, opened 2021-12-14
chore(tests): added security tests

laurentlp:
This is one of many PRs that adds security tests to our HTTP API

Closes DEV-2057


packages/server/src/api.ts
```diff
@@ -45,8 +45,14 @@ export class Api {
...
     this.root.use('/api', this.router)
-    this.router.use(cors())
-    this.router.use(express.json())
+    this.router.use(
+      cors({
+        origin: '*',
+        methods: ['GET', 'POST', 'DELETE'],
+        allowedHeaders: ['Content-Type', 'Authorization', 'Content-Length']
+      })
+    )
```
samuelmasse, line 54: What does this do?

packages/server/src/index.ts
```diff
@@ -17,6 +17,7 @@ const launch = async () => {
   await setupEnv()
 
   const router = express()
+  router.disable('x-powered-by')
```
samuelmasse, line 20: Why this change? Is this related to security?

packages/server/src/sync/api.ts
```diff
@@ -19,13 +20,18 @@ export class SyncApi {
     this.force = !!yn(process.env.SPINNED)
     this.schema = makeSyncRequestSchema(this.channels.list())
 
-    router.post('/sync', auth.public.auth(this.sync.bind(this)))
+    router
+      .route('/sync')
+      .post(auth.public.auth(this.sync.bind(this)))
+      .all(methodNotAllowed('POST'))
```
samuelmasse, line 26: If we're only going to implement methodNotAllowed for this route we might as well just not bother with it. It's not really additional security anyways.

packages/server/test/security/api.test.ts
```diff
@@ -0,0 +1,1005 @@
...
+
+        clients.first.userId = res.id
+
+        {
+          const res = await user(clients.second.clientId, clients.second.clientToken)
+
+          expect(res.id).not.toEqual(clients.first.userId)
+
+          clients.second.userId = res.id
+        }
```
samuelmasse, line 268: Why is there a random bracket here?

samuelmasse (approved):
Nice work! I have a few comments to clarify but overall these tests should prevent us from accidentally opening some of our routes to attacks.

I have a suggestion for another test : we could try sending payload of incorrect format, like url-econded

packages/server/src/api.ts
laurentlp: This simply strengthens the CORS for our API. It basically tells that allowed requests from any origin, that we only allow GET, POST, and DELETE, and some headers. This is mostly for requests coming from a browser but it still good practice to be as restrictive as possible.

packages/server/src/index.ts
laurentlp: Yes and no. It only removes the `x-powered-by: 'Express'` headers so that most people wouldn't necessarily know what framework powers our API. I know that we are an open-source project but again, it's only because it's a good practice thing. In our case it doesn't really have any negative nor positive effects.

packages/server/test/security/api.test.ts
laurentlp: To create another block of code. That way I can redefine a res constant without having to name it res2 or something

packages/server/src/sync/api.ts
laurentlp: Again as a good practice, we should implement this everywhere. Best scenario, it narrows the attack vector on our endpoints, and worst case, we have a middleware that brings no overhead!

PR #288, opened 2021-12-14
chore(repo): add fetch functions

samuelmasse:
We had way too many not nulls everywhere in the code, so this PR seperates all our `get` functions into `get` and `fetch`. A fetch functions is going to try to get the data, and return undefined if it doesn't find it. The get function will use the fetch function, and if the fetch function returns undefined, it will throw an error. This way we can use the get function in places in the code where we expect the get function to return a value, and remove all our not nulls. If for any reason the get function fails, it returns a clear message and prevents the request from running any longer with garbage data.

Closes DEV-2166

### DIFF
--- a/packages/server/src/clients/service.ts
+++ b/packages/server/src/clients/service.ts
@@ -56,7 +56,7 @@ export class ClientService extends Service {
     return client
   }
 
-  async getById(id: uuid): Promise<Client | undefined> {
+  async fetchById(id: uuid): Promise<Client | undefined> {
     const cached = this.cacheById.get(id)
     if (cached) {
       return cached
@@ -72,8 +72,16 @@ export class ClientService extends Service {
     }
   }
 
+  async getById(id: uuid): Promise<Client> {
+    const val = await this.fetchById(id)
+    if (!val) {
+      throw Error(`Client ${id} not found`)
+    }
+    return val
+  }
+
   async getByIdAndToken(id: uuid, token: string): Promise<Client | undefined> {
-    const client = await this.getById(id)
+    const client = await this.fetchById(id)
     if (!client) {
       return undefined
     }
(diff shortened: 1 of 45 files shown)


packages/server/src/clients/service.ts
```diff
@@ -56,7 +56,7 @@ export class ClientService extends Service {
     return client
   }
 
-  async getById(id: uuid): Promise<Client | undefined> {
+  async fetchById(id: uuid): Promise<Client | undefined> {
```
laurentlp, line 59: Private?

packages/server/src/messages/api.ts
```diff
@@ -152,12 +152,12 @@ export class MessageApi {
...
 
-    const message = await this.messages.get(id)
+    const message = await this.messages.fetch(id)
     if (!message) {
       return res.sendStatus(404)
     }
 
-    const conversation = await this.conversations.get(message.conversationId)
+    const conversation = await this.conversations.fetch(message.conversationId)
     if (!conversation || conversation.clientId !== req.client!.id) {
```
laurentlp, line 161:
```suggestion
    if (!conversation || conversation.clientId !== req.client.id) {
```

laurentlp (approved):

packages/server/src/messages/api.ts
samuelmasse: this is fixed

packages/server/src/clients/service.ts
samuelmasse: The fetch function can and is often used outside the service. It's not a public/private thing. The difference between the get and fetch function is intent. With the get function you indicate that the result must not be null, whereas with the fetch you indicate that you will handle the null value yourself.

PR #289, opened 2021-12-14
chore(tests): removed deprecation warnings and console

laurentlp:
This PR removes the deprecation warning for `rmdirSync`, bumps jest related dependencies, and removes the console.error from the components tests.


package.json
```diff
@@ -38,6 +38,7 @@
   "devDependencies": {
     "@types/jest": "^27.0.3",
     "@types/jest-dev-server": "^5.0.0",
+    "@types/node": "^16.11.13",
```
samuelmasse, line 41: Why is this here now?

samuelmasse (approved):
YES. The console.error was really annoying.

package.json
laurentlp: Because more than one package uses those typings, so it makes more sense that we put it in the root package.json

samuelmasse: I'd put them in the engine package instead. Rigth now the server package as mostly all dependencies, but I'll make a cleanup and put the correct deps in the engine and channels package

PR #290, opened 2021-12-14
fix(migration): fix alter table migration not working with sqlite

samuelmasse:
We simply disable foreign keys when running sqlite migrations


packages/server/src/conduits/events.ts
```diff
@@ -1,10 +1,9 @@
...
 
-// Note: event type ids should be random numbers between -9007199254740991 and 9007199254740991
 export enum ConduitEvents {
-  Created = -8735528850944253,
-  Deleting = -2992045965298849,
-  Updated = -4333216986594445
+  Created,
+  Deleting,
+  Updated
 }
```
samuelmasse, line 7: This is completely unrelated but these hardcoded ids don't have any use so I removed them

laurentlp (approved):
Tested and LGTM!

packages/server/src/build.ts
```diff
@@ -13,4 +13,4 @@ try {
   })
 } catch (e) {
   console.error("Couldn't get active branch", e)
-}
+}
```
laurentlp, line 16: Missing EOL

PR #291, opened 2021-12-15
chore(server): upgrade to yarn 3

samuelmasse:
Upgrades the server to yarn 3. At the moment parcel won't build, so we might need to upgrade to v2 as well.

Closes DEV-2164


laurentlp:
Will most probably require changes in the GitHub actions where were cached the node_modules

.github/workflows/build.yml
```diff
@@ -14,7 +14,7 @@ jobs:
           cache: 'yarn'
       - name: Fetch Node Packages
         run: |
-          yarn --frozen-lockfile
+          yarn --immutable
```
samuelmasse, line 17: Note : --frozen-lockfile doesn't exist anymore and was changes to --immutable with yarn 2

.vscode/extensions.json
```diff
@@ -0,0 +1,7 @@
+{
+  "recommendations": [
+    "arcanis.vscode-zipfs",
+    "dbaeumer.vscode-eslint",
+    "esbenp.prettier-vscode"
+  ]
+}
```
samuelmasse, line 7: When prompted by vscode to install these extensions you need to install them otherwise code lookup and other stuff just doesn't work.

.yarn/sdks/typescript/lib/tsc.js
```diff
@@ -0,0 +1,20 @@
...
+
+const relPnpApiPath = "../../../../.pnp.cjs";
+
+const absPnpApiPath = resolve(__dirname, relPnpApiPath);
+const absRequire = (createRequire || createRequireFromPath)(absPnpApiPath);
+
+if (existsSync(absPnpApiPath)) {
+  if (!process.versions.pnp) {
+    // Setup the environment to be able to require typescript/lib/tsc.js
+    require(absPnpApiPath).setup();
```
samuelmasse, line 15: Note : the code in the .yarn/sdks is library code, I didn't write any of this!

package.json
```diff
@@ -36,39 +36,43 @@
...
+    "eslint-import-resolver-node": "^0.3.6",
     "eslint-plugin-import": "^2.24.1",
     "eslint-plugin-jsdoc": "35.5.1",
     "fs-extra": "^10.0.0",
     "jest": "^27.4.5",
     "jest-dev-server": "^6.0.3",
-    "parcel": "1.12.3",
+    "jest-mock": "^27.4.2",
     "pkg": "^5.5.1",
+    "portfinder": "^1.0.28",
```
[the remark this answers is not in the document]
samuelmasse: Why not?

packages/components/src/renderer/custom.tsx
```diff
@@ -9,9 +9,10 @@ const checkError = (moduleInjector: Function, component: string, payload: string
   if (!moduleInjector) {
     return new Error(`${errorPrepend} could not get module injector`)
   }
-  if (!module) {
-    return new Error(`${errorPrepend} "module" is not defined in the payload`)
-  }
+  // TODO: what is this?
+  // if (!module) {
+  //   return new Error(`${errorPrepend} "module" is not defined in the payload`)
+  // }
```
samuelmasse, line 15: Any ideas?

package.json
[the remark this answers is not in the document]
laurentlp: Because only the server package uses it and already has it as its dependency (https://github.com/botpress/messaging/blob/sm-yarn-3/packages/server/package.json#L42)

packages/components/src/renderer/custom.tsx
laurentlp: I think `payload` should be renamed to `module` in the function's args as it would make much more sense (@charlescatta maybe you can add details here?)

But if I understand this piece of code correctly, when rendering custom content-types, we want to import the renderer from a certain module using the `InjectedModuleView`. This is the reason why we want to check if the module is specified, otherwise, it would fail to render.

laurentlp (approved):

PR #293, opened 2021-12-21
chore(dev): bump dependencies

laurentlp:
Pretty self-explanatory

Closes DEV-2156

Mainly: 
 - Moved some packages to their respective project (see engine's package.json)
 - Bump packages that could be upgraded without major changes in the codebase.


samuelmasse:
Most of these changes of moving deps to their correct packages is done in #291. I don't think we will merge this as it's going to have way to many conflicts with yarn 3

PR #294, opened 2021-12-21
chore(tests): added migration tests

laurentlp:
This PR adds tests to the migrations making sure that they work properly on both directions (`up` / `down`)

Closes DEV-2193


packages/server/test/migration/0.1.19-status.test.ts
```diff
@@ -0,0 +1,205 @@
...
+  let migration: MigrationService
+  let database: DatabaseService
+
+  beforeAll(async () => {
+    await setupApp()
+    migration = app.migration
+    database = app.database
+
+    migration.setupMigrations([StatusMigration])
+  })
```
samuelmasse, line 26: Setting up the migration service with a single migration will not work in the future if another migration affects this table. Consider this scenario : in the future the status table is removed completely in a migration that creates a new table called 'state'. When running this test, the down migration for the StatusMigration wouldn't work at all because the status table would not exist. It only works here because the version of the Server code produces a Database with the expected format for the test, but it won't be the case in the future.

A solution to this would be to adjust the database schema to be what we expect by doing an initial down migration with all the migrations to the target version, and after that calling `migration.setupMigrations([StatusMigration])` to test the StatusMigration exclusively.

samuelmasse (approved):
Good job! Like with the migrations, we don't want to use any services in the tests to avoid having to rewrite them all the time (except for Database and Migration service). I can see the tests were written without dependencies to the services so that's good. I wonder now if we could only setup the Database, Meta and Migration services to run these tests?

I noticed a flaw that might become a problem in the future when migrations have dependency on each other. We can merge this PR for now and fix these issues in a follow-up PR.

packages/server/test/migration/0.1.19-status.test.ts
samuelmasse: Another solution would be to not tests all the migrations independently at all, and start from the lastest one and down migrate to the first one, and then back up again with some jitter in between

laurentlp: Good point, we'll handle this once it becomes an issue!

packages/server/test/migration/0.1.19-status.test.ts
```diff
@@ -0,0 +1,205 @@
...
+
+  afterAll(async () => {
+    await app.destroy()
+    await database.destroy()
+  })
```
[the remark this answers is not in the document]
laurentlp: Good point, not needed!

PR #295, opened 2022-01-06
chore(server): integration tests

samuelmasse:
Adds integration tests some services

Closes DEV-2051


packages/server/test/integration/conduits.test.ts
```diff
@@ -0,0 +1,160 @@
...
+
+  beforeAll(async () => {
+    await setupApp()
+    conduits = app.conduits
+    querySpy = jest.spyOn(conduits as any, 'query')
+
+    state = {
+      provider: await app.providers.create(crypto.randomBytes(20).toString('hex'), false),
+      channel: app.channels.getByName('telegram'),
+      channelConfig: { botToken: 'blabla' }
```
samuelmasse, line 23: Replace these semi random string by a util function such as randomStr

packages/server/test/integration/clients.test.ts
```diff
@@ -13,7 +13,7 @@ describe('Clients', () => {
   beforeAll(async () => {
     await setupApp()
     clients = app.clients
-    querySpy = jest.spyOn(clients, 'query')
+    querySpy = jest.spyOn(clients as any, 'query')
```
laurentlp, line 16: PP: Cleaner IMHO

```suggestion
    querySpy = jest.spyOn<any>(clients, 'query')
```

packages/server/test/integration/providers.test.ts
```diff
@@ -12,7 +11,7 @@ describe('Providers', () => {
   beforeAll(async () => {
     await setupApp()
     providers = app.providers
-    querySpy = jest.spyOn(providers, 'query')
+    querySpy = jest.spyOn(providers as any, 'query')
```
laurentlp, line 14: also

packages/server/test/integration/health.test.ts
```diff
@@ -0,0 +1,97 @@
...
+
+  beforeEach(async () => {
+    app.caching.resetAll()
+  })
+
+  test('Register health event', async () => {
+    const data = { val: randStr() }
+    await health.register(state.conduit.id, HealthEventType.StartFailure, data)
+
+    state.eventData = data
```
laurentlp, line 43: Can you simply add something like `expect(data).toBeDefined()`

packages/server/test/integration/conduits.test.ts
```diff
@@ -0,0 +1,162 @@
...
+
+  test('Updating conduit config clears cache and persist changes', async () => {
+    const newConfig = { botToken: randStr() }
+    await conduits.updateConfig(state.conduit!.id, newConfig)
+    const calls = querySpy.mock.calls.length
+
+    const conduit = await conduits.get(state.conduit!.id)
+    expect(conduit).toEqual({ ...state.conduit, config: newConfig })
+    // TODO: doesn't work because this is reacted to in an event. Should events be disabled in tests?
+    // expect(querySpy).toHaveBeenCalledTimes(calls + 1)
```
laurentlp, line 101: You could maybe spy on the event dispatcher too then?

packages/server/test/integration/health.test.ts
samuelmasse: Data is defined earlier. The health.register function doesn't return anything

packages/server/test/integration/conversations.test.ts
```diff
@@ -0,0 +1,112 @@
...
+    const conversation = await conversations.get(state.conversation!.id)
+    expect(conversation).toEqual(state.conversation)
+    expect(querySpy).toHaveBeenCalledTimes(1)
+
+    for (let i = 0; i < 10; i++) {
+      const cached = await conversations.get(state.conversation!.id)
+      expect(cached).toEqual(state.conversation)
+    }
+
+    expect(querySpy).toHaveBeenCalledTimes(1)
```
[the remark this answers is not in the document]
samuelmasse: Yes. It is called 1 time in total for this test. Meaning that it is called once when we fetched the uncached value. At the end we verify that this number of times called is still 1

packages/server/test/integration/clients.test.ts
samuelmasse: You need more than one type argument to make this work. I that case I'd argue writing as any is less anoyying than writing jest.spyOn<any, any>(clients, 'query').

packages/server/test/integration/conduits.test.ts
samuelmasse: We don't really need to check that here anyways so I removed it.

packages/server/test/integration/health.test.ts
laurentlp: Yeah right, I misread the test, my bad!

laurentlp (approved):

PR #296, opened 2022-01-07
chore(repo): fix yarn 3 problems

samuelmasse:
Fixes problems brought by #291 

Closes DEV-2209


Dockerfile
```diff
@@ -25,10 +25,11 @@ COPY --from=build /messaging/packages/base/package.json packages/base/package.js
...
 COPY --from=build /messaging/package.json package.json
 COPY --from=build /messaging/yarn.lock yarn.lock
-COPY --from=build /messaging/.yarn .yarn
+COPY --from=build /messaging/.yarn/plugins .yarn/plugins
+COPY --from=build /messaging/.yarn/releases .yarn/releases
 COPY --from=build /messaging/.yarnrc.yml .yarnrc.yml
-COPY --from=build /messaging/.pnp.cjs .pnp.cjs
-COPY --from=build /messaging/.pnp.loader.mjs .pnp.loader.mjs
+
+RUN yarn workspaces focus --all --production
```
laurentlp, line 32: Equivalent of `yarn --production` (See: https://yarnpkg.com/getting-started/migration#renamed)

Dockerfile
```diff
@@ -25,10 +25,11 @@ COPY --from=build /messaging/packages/base/package.json packages/base/package.js
 
 COPY --from=build /messaging/package.json package.json
 COPY --from=build /messaging/yarn.lock yarn.lock
-COPY --from=build /messaging/.yarn .yarn
+COPY --from=build /messaging/.yarn/plugins/@yarnpkg/plugin-workspace-tools.cjs .yarn/plugins/@yarnpkg/plugin-workspace-tools.cjs
+COPY --from=build /messaging/.yarn/releases .yarn/releases
```
laurentlp, line 29: Here I only copied the plugin and yarn bin so since we don't require any other files from the `.yarn` folder. `node_modules` are installed on line 32.

package.json
```diff
@@ -39,14 +39,27 @@
...
+    "@parcel/transformer-typescript-tsc": "^2.1.1",
+    "@parcel/transformer-typescript-types": "^2.1.1",
     "@types/jest": "^27.0.3",
     "@types/jest-dev-server": "^5.0.0",
     "@types/node": "^16.11.13",
     "@types/uuid": "^8.3.3",
     "@typescript-eslint/eslint-plugin": "^4.29.3",
     "@typescript-eslint/parser": "^4.29.3",
     "bluebird-global": "^1.0.1",
-    "chalk": "^5.0.0",
```
laurentlp, line 49: chalk >= 5 only exports a ESM bundle which is not compatible with the way we import modules in the script folder (CommonJS).

scripts/release.js
```diff
@@ -15,10 +12,7 @@ yargs
     })
 
     try {
-      const version = (await fse.readJSON(path.join('package.json'))).version
-      const newVersion = semver.inc(version, argv.releaseType)
-
-      const bumpCommand = `version --new-version ${newVersion} --no-git-tag-version`
+      const bumpCommand = `version ${argv.releaseType}`
       await execute(`yarn ${bumpCommand}`, undefined, { silent: true })
```
laurentlp, line 16: Here I'm using the yarn `version` plugin to bump the packages versions (See: https://yarnpkg.com/cli/version)

Dockerfile
laurentlp: Edit: I had to copy all the plugins since yarn expects them all. Probably because the `.yarnrc` file lists them all.

laurentlp (approved):

PR #270, opened 2021-12-02
feat(channels): channels package

samuelmasse:
Moves all channels to their own package.

Each channel will have an api that will respect these standards : 

```ts
const telegram = new TelegramChannel()
await telegram.setup(app)

telegram.on('message', async (e) => {
  await telegram.send(e.scope, e.endpoint, { text: 'yoyo' })
})

await telegram.start('gggg', { botToken: 'YOUR_CONFIG' })
```

Closes DEV-2125

PR #297, opened 2022-01-10
fix(channels): readd clearing and fix auto start

samuelmasse:
This PR fixes some problems that were introduced by #270 

Notably the notion of keeping a lru of most used channels, and clearing the ones that aren't used was removed. Since this is a potential memory leak (although you'd really need to have a lot of channels), this PR fixes that. It also fixes channels auto starting and bypassing the normal start for a channels. So for example a channel could start without producing a health event.

Since these changes concerned the instance service, I took the time to refactor it into multiple sub services since the code was getting messy

Closes DEV-2206
Closes DEV-2212


samuelmasse:
Note: It's possible that a channel gets cleared up while having messages queued to be sent out. In that case we should wait before clearing the channel, otherwise there will be very rare and very hard to debug misses when sending out messages.

packages/server/test/integration/conduits.test.ts
```diff
@@ -41,6 +41,7 @@ describe('Conduits', () => {
     expect(conduit.config).toEqual(state.channelConfig)
 
     state.conduit = conduit
+    await app.instances.lifetimes.stop(conduit.id)
```
laurentlp, line 44: Why stop it?

laurentlp (commented):
Nice work as usual!!

Here are some missing parts:
 - Autostart (botpress -> messaging) missing.
 - Autostart callback needs to handle the case where a conduit is errored and does not start it.

packages/server/test/integration/conduits.test.ts
samuelmasse: Messes up test metrics because it sends some events

packages/channels/src/base/channel.ts
```diff
@@ -99,10 +106,14 @@ export abstract class ChannelTemplate<
     }
   }
 
-  autoStart(callback: (scope: string) => Promise<TConfig>) {
+  autoStart(callback: (scope: string) => Promise<void>) {
     this.service.autoStart(callback)
   }
 
+  stateManager(manager: ChannelStateManager): void {
```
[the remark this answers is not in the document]
samuelmasse: To be done in future pr

packages/channels/src/base/service.ts
```diff
@@ -49,13 +58,17 @@ export abstract class ChannelService<
...
-    await this.start(scope, await this.startCallback!(scope))
+    await this.startCallback!(scope)
   }
 
-  autoStart(callback: (scope: string) => Promise<TConfig>) {
+  autoStart(callback: (scope: string) => Promise<void>) {
     this.startCallback = callback
   }
 
+  stateManager(manager: ChannelStateManager) {
```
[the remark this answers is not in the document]
samuelmasse: Future pr

packages/channels/src/base/stream.ts
```diff
@@ -15,6 +15,8 @@ export abstract class ChannelStream<TService extends ChannelService<any, any>, T
   }
 
   protected async handleSend({ scope, endpoint, content }: ChannelSendEvent) {
+    await this.service.require(scope)
+
```
samuelmasse, line 19: Basically this was needed to be done to make the auto start from bp -> channel work

packages/server/src/instances/clearing/service.ts
```diff
@@ -53,7 +54,22 @@ export class InstanceClearingService extends Service {
...
+  private handleCacheDispose(key: string, value: any) {
+    if (this.destroyed) {
+      return
+    }
+
+    // if it's been deleted intentionally we assume disposing was done
+    if (this.statesDeleted) {
+      delete this.statesDeleted[key]
+      return
+    }
```
samuelmasse, line 66: I added this here because the dispose method from the lru gets called from values that are removed intentionnaly as well. This was causing two health events of type sleep to be emitted, so I put this safety here.

packages/server/src/instances/lifetime/service.ts
```diff
@@ -7,6 +8,8 @@ import { StatusService } from '../../status/service'
 import { InstanceLifetimeDispatcher, InstanceLifetimeDispatches } from './dispatch'
 import { InstanceLifetimeEmitter, InstanceLifetimeEvents, InstanceLifetimeWatcher } from './events'
 
+const MAX_ALLOWED_FAILURES = 5
```
laurentlp, line 11: This exact value is already defined in the monitoring service. Maybe we should consider exporting it instead of duplicating it? Unless we want to be able to use different values here than the monitoring service?

packages/server/src/sync/service.ts
```diff
@@ -115,6 +117,13 @@ export class SyncService extends Service {
 
         if (!_.isEqual(configWithoutEnabled, oldConduit.config)) {
           await this.conduits.updateConfig(oldConduit.id, configWithoutEnabled)
+        } else {
+          // updating the config will clear the number of errors.
+          // But if the config is identical we still want to clear it
+          const status = await this.status.fetch(oldConduit.id)
+          if ((status?.numberOfErrors || 0) > 0) {
```
laurentlp, line 124: PP: Since those two conditions are equal

```suggestion
          if (status?.numberOfErrors) {
```

laurentlp (approved):

PR #298, opened 2022-01-10
feat(client): client tokens

samuelmasse:
Changes how client tokens work to work the same way as user tokens.

Closes DEV-2205


packages/server/src/user-tokens/service.ts
```diff
@@ -94,12 +94,22 @@ export class UserTokenService extends Service {
...
+
     const userToken = await this.fetchById(id)
     if (!userToken) {
       return undefined
     }
 
+    if (userToken.userId !== userId) {
+      return undefined
+    }
+
```
samuelmasse, line 112: Fixed a huge security flaw. We didn't check if a user token belonged to that user! This needs to be added to the security tests. Otherwise another user can just use its own credentials to pretend to be any user of that belongs to the same clientId.

samuelmasse:
Not mergeable yet. We need to do a migration and also figure out how we will migrate the tokens in botpress, since this PR invalidates them.

packages/server/src/migrations/0.1.21-client-tokens.ts
```diff
@@ -0,0 +1,47 @@
...
+  async applied() {
+    return !(await this.trx.schema.hasColumn('msg_clients', 'token'))
+  }
+
+  async up() {
+    const clients = await this.trx('msg_clients')
+
+    for (const client of clients) {
+      await this.trx('msg_client_tokens').insert({ id: uuidv4(), clientId: client.id, token: client.token })
+    }
```
samuelmasse, line 24: It would be possible to replace the uuidv4() here by a deterministic algorithm that takes the token and creates a uuid. This way it would be possible to upgrade / downgrade the token from anywhere, without having to look at the db or adding an api point

samuelmasse: Actually nevermind there's no way for the other end to know what hash was produced, since it's salted

packages/server/src/migrations/0.1.21-client-tokens.ts
```diff
@@ -0,0 +1,58 @@
...
+    return !(await this.trx.schema.hasColumn('msg_clients', 'token'))
+  }
+
+  async up() {
+    await this.trx.schema.createTable('msg_client_tokens', (table) => {
+      table.uuid('id').primary()
+      table.uuid('clientId').references('id').inTable('msg_clients')
+      table.string('token').notNullable()
+      table.timestamp('expiry').nullable()
+    })
```
samuelmasse, line 25: I added the creation of the table in the migration here. This actually needs to happen in the migration, otherwise migrations might be unstable if we take the latest format of that table, since this format could change in the future. I also added the deletion of the table in the down migration.

laurentlp: IMHO, if decide to go against our standards here, we expose ourselves to future problems and inconsistencies across the repo. I still don't understand why you choose to put the table creation here considering that the "problem" in question is valid for all other tables too? Why only in the particular case? 

What do you mean by `migrations might be unstable if we take the latest format of that table`? If this format is to change in the future, it will be done by another migration that would implement an up and down function.

packages/server/src/migrations/0.1.21-client-tokens.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  async down() {
+    await this.trx.schema.alterTable('msg_clients', (table) => {
+      table.string('token').unique()
+    })
+
+    const clients = await this.trx('msg_clients')
+
+    for (const client of clients) {
+      const clientTokens = await this.trx('msg_client_tokens').where({ clientId: client.id })
```
laurentlp, line 45: Can clientTokens be empty?

packages/server/test/migration/0.1.20-fix-client-schema.test.ts
```diff
@@ -60,8 +60,7 @@ describe('0.1.20 - Fix Client Schema', () => {
     const id = uuid()
```
laurentlp, line 60: I understand why you did those changes, but since this test should be done on a DB that matches with the server v0.1.20, I think we should instead add a mechanism that easily allows us to roll back the DB to a certain version before running migration tests. This way, we won't have to make those changes and the tests would be more accurate. What do you think?

packages/server/test/security/api.test.ts
```diff
@@ -5,7 +5,7 @@ import { v4 as uuid } from 'uuid'
 import froth from './mocha-froth'
 
 const UUID_LENGTH = uuid().length
-const TOKEN_LENGTH = 88
+const TOKEN_LENGTH = 125
```
laurentlp, line 8: We should maybe consider exporting this value where it is defined in the sync schema instead.

packages/server/test/security/api.test.ts
```diff
@@ -162,7 +162,7 @@ describe('API', () => {
               token: expect.anything(),
               webhooks: expect.anything()
             })
-            expect(err.response?.data).toEqual('"token" length must be 88 characters long')
+            expect(err.response?.data).toEqual('"token" length must be 125 characters long')
```
laurentlp, line 165:
```suggestion
            expect(err.response?.data).toEqual(`"token" length must be ${TOKEN_LENGTH} characters long`)
```

packages/server/src/sync/schema.ts
```diff
@@ -16,7 +16,7 @@ export const makeSyncRequestSchema = (channels: Channel[]) => {
     channels: Joi.object(channelsSchema).allow(null),
     webhooks: Joi.array().items(SyncWebhookSchema).allow(null),
     id: Joi.string().guid().allow(null),
-    token: Joi.string().length(88).allow(null),
+    token: Joi.string().length(125).allow(null),
```
laurentlp, line 19: We should probably export those values by default so we can reuse them in our tests. What do you think?

packages/server/src/migrations/0.1.21-client-tokens.ts
samuelmasse: We don't create the table at two places, we create the table in the migrations for down/up compatibility, and we have code in the server that can create the database for the current version of the server. The migrations can't depend on the code that creates the table in the server since that code can change. So now any creation or modification of table is going to imply a migration, and the only use for the table creation code in the server is to create a new database. See #299

packages/server/src/client-tokens/service.ts
```diff
@@ -0,0 +1,125 @@
...
+      return undefined
+    }
+  }
+
+  async verifyToken(clientId: string, rawToken: string): Promise<ClientToken | undefined> {
+    if (!rawToken?.length) {
+      return undefined
+    }
+
+    const [id, token] = rawToken.split('.')
```
laurentlp, line 72: Can you put this separator in a constant so we can reuse it across services and in tests

packages/server/src/client-tokens/service.ts
```diff
@@ -0,0 +1,125 @@
...
+    const clientToken = await this.fetchById(id)
+    if (!clientToken) {
+      return undefined
+    }
+
+    if (clientToken.clientId !== clientId) {
+      return undefined
+    }
+
+    if (clientToken.expiry && Date.now() > clientToken.expiry.getTime()) {
```
laurentlp, line 87: Should we delete the token in this case?

packages/server/test/integration/client-tokens.test.ts
```diff
@@ -0,0 +1,140 @@
...
+import { app, randStr, setupApp } from './utils'
+
+describe('ClientTokens', () => {
+  let clientTokens: ClientTokenService
+  let querySpy: jest.SpyInstance
+  let state: { provider: Provider; client: Client; rawToken?: string; clientToken?: ClientToken }
+
+  beforeAll(async () => {
+    // This should be reset after those tests
+    process.env.ENABLE_EXPERIMENTAL_SOCKETS = '1'
```
laurentlp, line 15: I don't think there is anything wrong with enabling this feature across all integration tests.

packages/server/test/integration/client-tokens.test.ts
```diff
@@ -0,0 +1,140 @@
...
+    await app.destroy()
+  })
+
+  beforeEach(async () => {
+    app.caching.resetAll()
+  })
+
+  test('Create raw client token', async () => {
+    const rawToken = await clientTokens.generateToken()
+    expect(rawToken).toHaveLength(88)
```
laurentlp, line 40: Put the 88 in constant

laurentlp (commented):
Overall it LGTM! Brilliant job here!

packages/server/src/client-tokens/service.ts
samuelmasse: Yes we'll need a token janitor eventually but we don't care right now since the expiration feature isn't used.

packages/server/src/migrations/0.1.21-client-tokens.ts
samuelmasse: No why? It doesn't really make sense, you would just have a client that is inaccessible because it doesn't have any tokens. There's always going to be a master token that has all rigths.

packages/server/test/migration/0.1.20-fix-client-schema.test.ts
samuelmasse: Agreed. I'll make changes to how migration tests are done in #299. I think it's ok to merge this for now

packages/server/test/security/api.test.ts
samuelmasse: I'll actually remove this limitation. With the legacy handling tokens can be of more than one size.

laurentlp (approved):

PR #299, opened 2022-01-12
feat(migrations): improve migration system

samuelmasse:
Assures that the database version is persisted in a transaction that also contains any changes to the database schema. Tables are now created initially in a transaction all at the same time. Changes the status migration to create the table in the correct state after the migration has run (this is necessary to be future proof, in case a migration in the future needs to modify the schema again).

When adding a table, it is now required to add a migration (not only when modifying an existing table). This is required because not doing so might lead to problems in the future where running multiple migrations in a row could lead to corrupt database state, since a table that is needed does not exist in the expected format. We also can't allow the server to create its tables before running the migrations, because doing so could also lead to to problems in the future where a migration might expect to migrate a table as it was in a specific version but the server creates a table of a newer format that the migration doesn't expect.

Closes DEV-2216


packages/server/src/migrations/0.0.1-init.ts
```diff
@@ -0,0 +1,179 @@
...
+
+    await this.trx.schema.createTable('msg_kvs', (table) => {
+      table.uuid('id').primary()
+      table.string('key').unique().notNullable()
+      table.jsonb('value').notNullable()
+    })
+
+    await this.trx.schema.createTable('msg_channels', (table) => {
+      table.uuid('id').primary()
+      table.string('name').unique().notNullable()
```
samuelmasse, line 32: We need this massive initial migration to be coherent across the system. That way you can do something like `migrate up -t 0.0.1` on a messaging server that hasn't created its database yet, and this will create a database compatible for version 0.0.1. Same goes for any other version as well. So now our database tests will be able to compare sequential migration runs to known schemas, and individual migration tests will be able to demand the database version they need by migrating to it from version 0.0.1 before doing their tests

laurentlp (commented):
Small comments otherwise looks fine to me!

laurentlp (approved):

PR #304, opened 2022-01-14
chore(server): rewrite http client & split sync

samuelmasse:
Rewrites the messaging http client to a similar implementation of the socket client

This is the new implementation
```ts
const client = new MessagingClient({
  url: 'my-url',
  clientId: 'my-client-id', 
  clientToken: 'my-client-token', 
  webhookToken: 'my-webhook-token'
})
client.setup(router)

client.on('message', async (e) => {
  await client.createMessage(e.conversationId, undefined, { text: 'yoyo' })
})
```
Under the hood this MessagingClient uses the MessagingChannel and initializes one client in it
```ts
const messaging = new MessagingChannel({ url: 'my-url' })
await messaging.setup(router)

messaging.on('message', async (clientId, e) => {
  await messaging.createMessage(clientId, e.conversationId, undefined, { text: 'yoyo' })
})

await messaging.start('my-client-id', { clientToken: 'my-client-token', webhookToken: 'my-webhook-token' })
```
Notice how it is very similar to the new channel implemenation. It makes sense, since the role of messaging after all is to unify all these channels.
```ts
const telegram = new TelegramChannel()
await telegram.setup(router)

telegram.on('message', async (e) => {
  await telegram.send(e.scope, e.endpoint, { text: 'yoyo' })
})

await telegram.start('gggg', { botToken: 'YOUR_CONFIG' })
```
It's also pretty close to the socket client 
```ts
const socket = new MessagingSocket({
  clientId: 'd0679b50-4c7d-4afe-a10b-fc98e06e318a'
})

socket.on('message', (message) => {
  console.log('Bot response :', message.payload.text)
})

await socket.connect()

await socket.sendText('Hello there!')
```

Closes DEV-2219
Closes DEV-2073


packages/client/src/emitter.ts
```diff
@@ -0,0 +1,56 @@
...
+        await listener(clientId, arg)
+      }
+      return true
+    } else {
+      return false
+    }
+  }
+}
+
+export class ProtectedEmitter<T extends { [key: string]: any }> {
```
laurentlp, line 31: Not sure I understand what this is for? Why create two implementations of an Emitter class?

laurentlp (approved):
Small question otherwise LGTM! Awesome work!!!

packages/client/src/emitter.ts
samuelmasse: One has two arguments it sends out with events. The other only has one.

samuelmasse: This is to handle clientId as the first argument of callbacks

PR #306, opened 2022-01-14
chore(repo): etc folder

samuelmasse:


.npmrc
```diff
@@ -1,2 +0,0 @@
-
-engine-strict=true
```
laurentlp, line 2: This file was use to enforce the version of NodeJS we use for this project. Does it still force us to use the engine version specified in the package.json event without this file?

package.json
```diff
@@ -105,8 +106,11 @@
     "@botpress/chat-skin": "@botpress/chat-skin/src"
   },
   "dependenciesMeta": {
-    "@blueprintjs/core@3.52.0": {
-      "unplugged": true
+    "core-js@3.20.0": {
+      "built": false
+    },
+    "core-js-pure@3.20.0": {
+      "built": false
```
laurentlp, line 113: Why not build them?

packages/board/package.json
```diff
@@ -9,10 +9,14 @@
...
   "devDependencies": {
+    "@parcel/config-default": "^2.1.1",
+    "@parcel/transformer-typescript-tsc": "^2.1.1",
     "@types/react": "^17.0.38",
-    "@types/react-dom": "^17.0.11"
+    "@types/react-dom": "^17.0.11",
+    "typescript": ">=3.0.0"
   },
   "dependencies": {
+    "@botpress/messaging-base": "0.0.1",
```
laurentlp, line 19: I think that with every local packages, we should use the `*` instead of a fixed version as we always want to use the one provided locally!

package.json
```diff
@@ -7,22 +7,22 @@
   "scripts": {
     "eslint": "eslint packages/ --ext .ts",
     "prettier": "prettier --check './packages/**/*.ts' '!**/*.d.ts'",
-    "clean": "rimraf packages/*/dist bin",
-    "reset": "yarn clean && rimraf .parcel-cache packages/*/node_modules && rimraf node_modules",
-    "package": "yarn clean && yarn build && node ./scripts/package",
-    "release": "yarn node ./scripts/release",
+    "clean": "mkdirp packages/base/dist && rimraf packages/*/dist bin etc/jest/.test-data",
```
laurentlp, line 10: You create a dist folder inside the base package and then do an `rm -rf` for every dist folder under every package. Is it because otherwise, `rimraf` would raise an error saying that there is no dist folder in the base package?

package.json
```diff
@@ -7,22 +7,22 @@
...
+    "reset": "yarn clean && mkdirp packages/base/node_modules && rimraf .parcel-cache packages/*/node_modules node_modules .yarn/cache .yarn/install-state.gz .yarn/unplugged .pnp.cjs .pnp.loader.mjs",
```
laurentlp, line 11: I think we shouldn't keep anything related to yarn PnP till we give it another go

laurentlp (commented):
Small comments other LGTM!

package.json
samuelmasse: It just says that there is no match for packages/*/dist

package.json
samuelmasse: Everything works for PnP except pkg so we can use PnP to sort out package deps. In fact we can put the package in PnP when developping we just need node_modules to create the binary.

package.json
samuelmasse: When using PnP, this adds 7000 files to the repo, out of a total of 10000 that is produce, so it reduces the amount of files to 3000 afterwards. I haven't seen anything not working from not building those. I think storybook uses them

packages/board/package.json
samuelmasse: Ok but we can make this change in another PR. This is not consistent across the codebase

.npmrc
samuelmasse: Don't know but it seems really random to have a npmrc file when we don't use npm. There's probably a yarn option for that.

package.json
laurentlp: Can you put this on another branch that we can then use to give PnP another go? That way, we won't pollute the repo with code/configs that are currently unused or unavailable.

package.json
laurentlp: Please let not do this as it will create a mess of configs and be so hard to debug in case we have issues.

.npmrc
laurentlp: Note: With yarn 2+, we have to install this plugin: https://github.com/devoto13/yarn-plugin-engines

laurentlp (approved):

PR #308, opened 2022-01-17
chore(dev): bump parcel

laurentlp:
Bumps parcel to 2.2.0 which only contains a security fix and a new feature (see: https://github.com/parcel-bundler/parcel/releases) and some patches for other dependencies.


samuelmasse (approved):
LGTM

PR #309, opened 2022-01-17
chore(dev): bump pkg

laurentlp:
Bumps `pkg` to `5.2.2` which now uses `pkg-fetch` v3.2.6 which installs `NodeJS` v16.13.2.


samuelmasse (approved):
LGTM

PR #311, opened 2022-01-17
chore(deps): always use local packages

laurentlp:
We use the `*` wildcard to make sure that we always use the local version of our packages.

_Note: I know that we will have multiple versions of the channel package, we'll only have to adapt our dependencies when we are there._

### DIFF
--- a/packages/inject/package.json
+++ b/packages/inject/package.json
@@ -17,7 +17,7 @@
     "@types/react-dom": "^17.0.11"
   },
   "dependencies": {
-    "@botpress/webchat": "0.0.1",
+    "@botpress/webchat": "*",
     "react": "^17.0.2",
     "react-dom": "^17.0.2"
   }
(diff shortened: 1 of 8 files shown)


packages/inject/package.json
```diff
@@ -17,7 +17,7 @@
     "@types/react-dom": "^17.0.11"
   },
   "dependencies": {
-    "@botpress/webchat": "0.0.1",
+    "@botpress/webchat": "*",
```
samuelmasse, line 20: Inject is a published package so shouldn't use * either

packages/webchat/package.json
```diff
@@ -24,7 +24,7 @@
   "dependencies": {
     "@blueprintjs/core": "^3.23.1",
     "@botpress/messaging-components": "*",
-    "@botpress/messaging-socket": "0.0.2",
+    "@botpress/messaging-socket": "*",
```
samuelmasse, line 27: Webchat is a published package so no * either

samuelmasse (approved):
There's to more * that shoudn't be there, otherwise LGTM

PR #312, opened 2022-01-17
chore(ci_cd): run tests both on sqlite and postgresql

laurentlp:
This PR edits the test action to include tests on PostgreSQL so that we cover running tests on all possible configurations.

Closes DEV-2157


.github/workflows/tests.yml
```diff
@@ -22,71 +22,139 @@ jobs:
...
           yarn --immutable
+      - name: Await PostgreSQL
+        if: ${{ matrix.database == 'postgres' }}
+        timeout-minutes: 2
+        run: until pg_isready -h localhost; do sleep 1; done
       - name: Run tests
         run: |
           yarn test:int --coverage --verbose
+        env:
+          DATABASE_URL: ${{ matrix.database == 'postgres' && 'postgres://postgres:postgres@localhost:5432/postgres' || '' }}
```
samuelmasse, line 56: Why the `|| ''` at the end there?

packages/engine/src/database/service.ts
```diff
@@ -76,6 +77,19 @@ export class DatabaseService extends Service {
...
+        const trx = await this.knex.transaction()
+        for (const table of this.tables.reverse()) {
+          await trx.schema.dropTable(table.id)
+          this.logger.debug(`Dropped table '${table.id}'`)
+        }
+        await trx.commit()
+      } catch (e) {
+        this.logger.error(e, 'Failed to destroy transient database')
+      }
+    }
```
samuelmasse, line 91: Here's a new setting. You can set your DB to be transient so it's cleared when the server closes.

packages/engine/src/global.ts
```diff
@@ -10,6 +10,8 @@ export interface EngineEnv {
   // database
   DATABASE_URL?: string
   DATABASE_POOL?: string
+  DATABASE_SUFFIX?: string
+  DATABASE_TRANSIENT?: string
```
samuelmasse, line 14: DATABASE_SUFFIX will be added at the end of table names

test/jest.e2e.setup.ts
```diff
@@ -1,14 +1,20 @@
...
 const setup = async () => {
   process.env.SKIP_LOAD_ENV = 'true'
   process.env.ENABLE_EXPERIMENTAL_SOCKETS = 'true'
-  process.env.DATABASE_URL = path.join(__dirname, '.test-data', `${uuidv4()}.sqlite`)
+  process.env.DATABASE_URL = process.env.DATABASE_URL || path.join(__dirname, '.test-data', `${uuidv4()}.sqlite`)
+
+  if (process.env.DATABASE_URL.startsWith('postgres')) {
+    process.env.DATABASE_SUFFIX = `__${randomLetters(8)}`
+    process.env.DATABASE_TRANSIENT = 'true'
+  }
```
samuelmasse, line 17: DATABASE_TRANSIENT doesn't work for e2e tests because jest doesn't let the server shut down completely. We'll need to look into a way of letting the dev server shutdown completely before killing the process. The database connection even stays alive.

samuelmasse: Fixed

.github/workflows/tests.yml
laurentlp: It's the equivalent of a ternary operator meaning:

`condition && true || false` instead of the classic `condition ? true : false `

laurentlp: Otherwise, I think it would set the value of `DATABASE_URL` to `false`

packages/engine/src/database/service.ts
laurentlp: Nice! But, I never know, if there is an error during the commit call, is rollback automatically called or we should rather call it explicitly in the catch block?

packages/engine/test/unit/database.test.ts
```diff
@@ -211,6 +211,10 @@ describe('DatabaseService', () => {
   })
 
   describe('SQLite', () => {
+    beforeEach(() => {
+      process.env.DATABASE_URL = undefined
```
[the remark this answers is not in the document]
samuelmasse: Otherwise it sets up a postgres database for some reason.

packages/engine/src/database/service.ts
samuelmasse: I know that changes aren't applied unless commit is called, but not calling rollback explicitly might cause the connection to fail to close. I'll add the rollback in the catch block

packages/server/src/health/table.ts
```diff
@@ -1,14 +1,14 @@
-import { Table } from '@botpress/messaging-engine'
+import { getTableId, Table } from '@botpress/messaging-engine'
 import { Knex } from 'knex'
 
 export class HealthTable extends Table {
-  get id() {
-    return 'msg_health'
+  get name() {
+    return getTableId('msg_health')
```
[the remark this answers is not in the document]
samuelmasse: Good catch that's completely wrong

packages/server/test/migration/0.1.19-status.test.ts
```diff
@@ -126,7 +127,7 @@ describe('0.1.19 - Status', () => {
         providerId: expect.anything(),
         config: expect.anything(),
         channelId: TELEGRAM_CHANNEL_ID,
-        initialized: Number(true)
+        initialized: expect.anything()
```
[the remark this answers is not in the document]
samuelmasse: initialized is a timestamp, not a boolean, so expecting 1 here doesn't work. I'm not sure why it worked for sqlite.

test/jest.e2e.setup.ts
```diff
@@ -21,4 +27,13 @@ const setup = async () => {
   })
 }
 
+const randomLetters = (length: number) => {
```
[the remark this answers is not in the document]
samuelmasse: Whatever it's not a complex function, and it would be annoying to try to share code at the two very different places it's used.

packages/engine/test/unit/database.test.ts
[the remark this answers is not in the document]
laurentlp: Weird considering that knex is mocked a few lines above

packages/server/test/integration/utils/index.ts
```diff
@@ -8,14 +8,33 @@ export let app: App
...
 
   app = new App()
   await app.setup()
   await app.postSetup()
   return app
 }
 
+export const destroyApp = async () => {
+  await app.destroy()
+}
```
laurentlp, line 27: This does not seem to be used anywhere!

samuelmasse: Good catch I did this before making DATABASE_TRANSIENT

samuelmasse (approved):

PR #313, opened 2022-01-17
chore(sync): split sync into two routes

samuelmasse:
There is now a /admin/client route that serves as the route to create a messaging client. The sync api no longer creates clients.

Closes DEV-2227


samuelmasse:
Note: ADMIN_KEY needs to be added to authenticate admin routes.

packages/server/src/base/api-manager.ts
```diff
@@ -33,3 +33,33 @@ export class ApiManager {
...
+  }
+
+  delete(path: string, schema: Joi.ObjectSchema<any>, fn: Middleware<Request>) {
+    this.use('delete', path, schema, fn)
+  }
+
+  use(type: 'post' | 'get' | 'delete', path: string, schema: Joi.ObjectSchema<any>, fn: Middleware<Request>) {
+    this.router[type](
+      path,
+      this.auth.public.auth(async (req, res) => {
```
laurentlp, line 55: Shouldn't we rather use something like `this.auth.admin.auth` instead?

packages/server/src/clients/api.ts
```diff
@@ -0,0 +1,103 @@
...
+    const client = await this.clients.create(provider.id, clientId)
+
+    const rawToken = await this.clientTokens.generateToken()
+    const clientToken = await this.clientTokens.create(client.id, rawToken, undefined)
+
+    res.status(201).send({ id: client.id, token: `${clientToken.id}.${rawToken}` })
+  }
+
+  async sync(req: Request, res: Response) {
+    this.distributed.using(`lock_dyn_client_sync::${req.body.name}`, async () => {
```
laurentlp, line 39:
```suggestion
    this.distributed.using(`lock_dyn_admin_client_sync::${req.body.name}`, async () => {
```

packages/server/src/clients/api.ts
```diff
@@ -0,0 +1,103 @@
...
+
+        await this.clients.updateProvider(client.id, provider.id)
+        client = await this.clients.getById(client.id)
+      }
+    }
+
+    if (!token || !(await this.clientTokens.verifyToken(client.id, token))) {
+      const rawToken = await this.clientTokens.generateToken()
+      const clientToken = await this.clientTokens.create(client.id, rawToken, undefined)
+      token = `${clientToken.id}.${rawToken}`
```
laurentlp, line 89: We should really consider putting this into some util function!

packages/server/src/sync/api.ts
```diff
@@ -1,69 +1,38 @@
...
-    const { error, value } = this.schema.validate(bodyWithoutEnabled)
-    if (error) {
-      return res.status(400).send(error.message)
-    }
+    const result = await this.syncs.sync(req.clientId, value)
+    res.send(result)
+  }
 
-    const sync = value as SyncRequest
+  private removeDisabledChannels(body: any) {
```
laurentlp, line 27: I would be easy here to use some kind of interface instead of any since all channels share the enable property!

packages/server/src/sync/schema.ts
```diff
@@ -12,11 +13,10 @@ export const makeSyncRequestSchema = (channels: Channel[]) => {
...
-    webhooks: Joi.array().items(SyncWebhookSchema).allow(null),
-    id: Joi.string().guid().allow(null),
-    token: Joi.string().max(125).allow(null),
-    name: Joi.string().max(50).allow(null)
+  return ReqSchema({
+    body: {
+      channels: Joi.object(channelsSchema).allow(null),
+      webhooks: Joi.array().items(SyncWebhookSchema).allow(null)
+    }
   }).options({ stripUnknown: true })
```
laurentlp, line 21: I think we can now remove this black magic and force users to send proper data. What do you think?

packages/server/test/security/api.test.ts
```diff
@@ -49,32 +50,58 @@ const shouldFail = async (func: Function, onError: (err: AxiosError) => void) =>
...
 describe('API', () => {
-  const sync = async (data?: SyncRequest, config?: AxiosRequestConfig) => {
-    const client = http()
+  describe('Admin', () => {
+    const createClient = async () => {
+      const client = http()
+      const res = await client.post('/api/admin/clients')
 
-    const res = await client.post<SyncResult>('/api/sync', data, config)
+      expect(res.data).toEqual({ id: expect.anything(), token: expect.anything() })
```
laurentlp, line 58: We could consider, in the future, implementing some kind of expect.uuid util. What do you think?

packages/server/test/security/api.test.ts
```diff
@@ -49,32 +50,58 @@ const shouldFail = async (func: Function, onError: (err: AxiosError) => void) =>
...
+
+      const res = await client.post<SyncResult>('/api/sync', data || {}, config)
+
+      expect(res.data).toEqual({ webhooks: expect.anything() })
+      expect(res.status).toEqual(200)
+
+      return res.data
+    }
+
     test('Should not allow any other methods than POST and OPTIONS', async () => {
```
laurentlp, line 94: I think we can remove this test as the response is now 404 when you query this endpoint with an unknown method

packages/server/test/security/api.test.ts
```diff
@@ -89,63 +116,46 @@ describe('API', () => {
       }
     })
 
-    test('Should allow anyone to make a sync request', async () => {
```
laurentlp, line 92: Maybe we could convert this test into `Should **not** allow everyone to make a sync request`

packages/server/src/base/api-manager.ts
samuelmasse: Yup added that. There is now an ADMIN_KEY env variable that protects these routes

packages/server/src/sync/api.ts
samuelmasse: I use any here since there can techically be anything in there, since we use the stripUnknown feature.

packages/server/src/sync/schema.ts
samuelmasse: We can't remove that unless we do a migration on botpress' side. What we can do is disable this when messaging is not spinned

packages/server/test/security/api.test.ts
samuelmasse: Could be useful. We could do that when revisiting the tests in another PR

packages/server/test/security/api.test.ts
samuelmasse: I added a test for that

packages/server/src/clients/api.ts
samuelmasse: For a future PR

packages/server/src/sync/schema.ts
laurentlp: Well plan for future us ;)

packages/client/src/base.ts
```diff
@@ -14,12 +14,14 @@ export abstract class MessagingChannelBase extends Emitter<{
   protected readonly http: AxiosInstance
   protected auths: { [clientId: uuid]: MessagingClientAuth } = {}
   protected headers: { [clientId: uuid]: any } = {}
+  protected adminHeader: any
```
laurentlp, line 17:
```suggestion
  protected adminHeader: { 'x-bp-messaging-admin-key'?: string }
```

packages/socket/test/e2e/socket.test.ts
```diff
@@ -12,7 +12,8 @@ const createClient = async () => {
     hostname: url.hostname,
     port: url.port,
     path: '/api/admin/clients',
-    method: 'POST'
+    method: 'POST',
+    headers: { 'x-bp-messaging-admin-key': 'admin123' }
```
laurentlp, line 16:
```suggestion
    headers: { 'x-bp-messaging-admin-key': process.env.ADMIN_KEY }
```

packages/server/test/security/api.test.ts
```diff
@@ -51,9 +51,20 @@ const shouldFail = async (func: Function, onError: (err: AxiosError) => void) =>
 
 describe('API', () => {
   describe('Admin', () => {
-    const createClient = async () => {
+    const createClient = async (override?: { adminKey?: string }) => {
+      let options: any = {
+        headers: { 'x-bp-messaging-admin-key': 'admin123' }
```
laurentlp, line 56:
```suggestion
        headers: { 'x-bp-messaging-admin-key': process.env.ADMIN_KEY }
```

packages/client/test/e2e/client.test.ts
```diff
@@ -28,7 +28,7 @@ describe('Http Client', () => {
     webhooks?: SyncWebhook[]
   } = {}
   const url = 'http://localhost:3100'
-  const adminClient = new MessagingChannel({ url })
+  const adminClient = new MessagingChannel({ url, adminKey: 'admin123' })
```
laurentlp, line 31:
```suggestion
  const adminClient = new MessagingChannel({ url, adminKey: process.env.ADMIN_KEY })
```

packages/client/src/base.ts
samuelmasse: Looks overkill to my. It's just an object of headers, you can have anything in there. It's specified just below what we put here.

laurentlp (approved):

PR #315, opened 2022-01-18
chore(server): v0.2.0

laurentlp:
Release new minor after changes in the client and improvements in our test suites!


.yarn/versions/6c73f69f.yml
```diff
@@ -0,0 +1,2 @@
+undecided:
+  - "@botpress/messaging"
```
samuelmasse, line 2: What's that??

PR #320, opened 2022-01-19
chore(server): v0.1.23

samuelmasse:
Release to fix #319. Merge after #319 and #318 is merged.


laurentlp (changes requested):
The release should be created after those PRs are merged this way, it will generate a proper changelog

PR #321, opened 2022-01-25
chore(channels): remove legacy channels from code

samuelmasse:
This PR removes the legacy channel implementations from the codebase, and adds a dependency to that code using npm. This is the start of the channels v2 implementation


packages/channels/example/app.ts
```diff
@@ -31,8 +16,18 @@ export class App {
...
       const respond = async () => {
         try {
-          for (const payload of payloads) {
-            await channel.send(scope, endpoint, payload)
+          if (content.type === 'say_something') {
+            await channel.send(scope, endpoint, { type: 'text', text: `Say Something: ${content.text}` })
+          } else if (content.type === 'postback') {
+            await channel.send(scope, endpoint, { type: 'text', text: `Postback: ${content.payload}` })
+          } else if (content.type === 'quick_reply') {
+            await channel.send(scope, endpoint, { type: 'text', text: `Quick Reply: ${content.text}` })
```
samuelmasse, line 24: This file basically removes the deleted channel from the channel testing server

packages/channels/example/payloads.json
```diff
@@ -80,12 +86,55 @@
...
     ]
   },
+  {
+    "type": "file",
+    "title": "here is a zip of the 1.0.0 messaging codebase",
+    "file": "https://github.com/botpress/messaging/archive/refs/tags/v1.0.0.zip",
+    "typing": true
+  },
+  {
+    "type": "video",
```
samuelmasse, line 96: Added a few payloads here to be able to test more content types

packages/channels/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@botpress/messaging-channels",
-  "version": "0.1.0",
+  "version": "0.2.0",
   "main": "dist/index.js",
```
samuelmasse, line 4: We'll use minor version 2 for temporary development until we release 1.0.0

packages/channels/src/base/api.ts
```diff
@@ -14,19 +12,6 @@ export class ChannelApi<TService extends ChannelService<any, any>> {
...
-    if (yn(process.env.SPINNED)) {
-      const externalUrl = await this.urlCallback!(scope)
-
-      this.service.logger?.info(
-        `[${scope}] ${clc.bold(name.charAt(0).toUpperCase() + name.slice(1))}${
-          path ? ' ' + path : ''
-        } webhook ${clc.blackBright(`${externalUrl}/${name}${path ? `/${path}` : ''}`)}`
-      )
-    }
-  }
```
samuelmasse, line 29: I'm removing this from here. I will still work with old channels but not the new ones. I suggest we change it so botpress prints this instead like we talked about

packages/channels/src/base/service.ts
```diff
@@ -112,9 +112,6 @@ export abstract class ChannelService<
     if (index) {
       const key = this.getIndexCacheKey(scope, identity, sender)
       const options = this.cacheIndexResponses.get(key)
-
-      this.cacheIndexResponses.del(key)
-
```
samuelmasse, line 117: Removing this since it makes sense that when you get a carousel in twilio, and there is a list of choices to make, you could make more than once choices successively. That line prevented you from doing that

packages/server/package.json
```diff
@@ -24,6 +24,7 @@
   "dependencies": {
     "@botpress/messaging-base": "*",
     "@botpress/messaging-channels": "*",
+    "@botpress/messaging-channels-legacy": "npm:@botpress/messaging-channels@0.1.0",
     "@botpress/messaging-engine": "*",
```
samuelmasse, line 28: That's how we are going to use two versions of each channel at the same time using this dependency alias

packages/server/src/rewire.ts
```diff
@@ -5,7 +5,11 @@ if (process.env.TS_NODE_DEV) {
...
 
   const rewire = function (this: (id: string) => any, mod: string) {
-    if (mod.startsWith('@botpress')) {
+    if (
+      mod.startsWith('@botpress/messaging') &&
+      // we don't want to rewire the legacy channels because we get them from npm
+      mod !== '@botpress/messaging-channels-legacy'
+    ) {
       return originalRequire.apply(this, [mod + '/src'])
     }
```
samuelmasse, line 14: Had to add this otherwise ts-node would try to run the legacy channel code from source

laurentlp (approved):

PR #322, opened 2022-01-25
feat(telegram): telegram v1

samuelmasse:
Rewrite of telegram channel with telegraf v4.6.0 (upgraded from the old version that used 3.27.1)

Closes DEV-2251


samuelmasse:
Ready for review but I need to add the handling the version management in the backend for channels. Right now you can use this new telegram implementation as a drop in for the previous one, but we'll have it as a different version

packages/channels/src/telegram/api.ts
```diff
@@ -0,0 +1,97 @@
...
+      await this.service.receive(scope, this.extractEndpoint(ctx), { type: 'text', text: ctx.message.text })
+    }
+  }
+
+  private async handleTelegrafCallbackQuery(
+    scope: string,
+    ctx: NarrowedContext<Context<Update>, Update.CallbackQueryUpdate>
+  ) {
+    if ('data' in ctx.callbackQuery) {
+      const data = ctx.callbackQuery.data
```
samuelmasse, line 62: As I think you've seen as well that's the suggested way of making the typings work as per telegraf documentation. It's a bit strange but whatever

packages/channels/src/telegram/api.ts
```diff
@@ -0,0 +1,97 @@
...
+    return (ctx: any) => {
+      fn(scope, ctx).catch((e) => {
+        this.service.logger?.error('Error occurred in telegram callback', e)
+      })
+    }
+  }
+
+  private useWebhook() {
+    // TODO: remove this dependency on server env vars
+    return !yn(process.env.SPINNED) || yn(process.env.CLUSTER_ENABLED)
```
samuelmasse, line 95: TODO: need to do this. Should be a allowSockets configuration or something like that.

packages/channels/src/telegram/channel.ts
```diff
@@ -0,0 +1,23 @@
...
+  get meta() {
+    return {
+      // TODO: change the meta here
+      id: '0198f4f5-6100-4549-92e5-da6cc31b4ad1',
+      name: 'telegram',
+      schema: TelegramConfigSchema,
+      initiable: true,
+      lazy: true
+    }
+  }
```
samuelmasse, line 17: I left it like the same metadata of the old telegram channel for now, so you can switch between the old and new implementation. Needs to be changed before merging this

packages/channels/src/telegram/config.ts
```diff
@@ -0,0 +1,9 @@
+import Joi from 'joi'
+
+export interface TelegramConfig {
+  botToken: string
+}
+
+export const TelegramConfigSchema = Joi.object({
+  botToken: Joi.string().required()
+}).options({ stripUnknown: true })
```
samuelmasse, line 9: It's likely we'll change the names of these config options to standarize it accross channels. For now we can merge this PR with that and standardize it later.

packages/channels/src/telegram/README.md
```diff
@@ -0,0 +1,28 @@
...
+| ------------- | :------: |
+| Text          |    ✅    |
+| Postback      |    ✅    |
+| Say Something |    ✅    |
+| Voice         |    ❌    |
+| Image         |    ❌    |
+| File          |    ❌    |
+| Audio         |    ❌    |
+| Video         |    ❌    |
+| Location      |    ❌    |
```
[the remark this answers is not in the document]
samuelmasse: Also added quick_reply for anything that is sent by pressing a button instead of writing text (like single-choice or dropdown).

samuelmasse:
TODO: add a way to disable automatic webhook creation. This should be an option in all channels

laurentlp (approved):

PR #326, opened 2022-01-27
fix(conversations): fix ordering

samuelmasse:
I messed up the conversation listing in the conversation service when I added added the getTableId function. Basically the conversations were not listed in order of most recently used anymore, so I fixed it and added a test to prevent it from not working again


samuelmasse:
@laurentlp we'll need to release 1.0.1 with this fix as this is pretty major

laurentlp (approved):
Nice catch! LGTM!

PR #328, opened 2022-01-27
feat(channels): channel versioning

samuelmasse:
Adds versioning to our channels. It's not possible for a single clientId to have more than one version of the same channel configured at the same time (by design).

Closes DEV-2261

### DIFF
--- a/packages/channels/src/base/meta.ts
+++ b/packages/channels/src/base/meta.ts
@@ -3,7 +3,8 @@ import Joi from 'joi'
 export interface ChannelMeta {
   id: string
   name: string
+  version: string
   initiable: boolean
   lazy: boolean
-  schema: Joi.ObjectSchema
+  schema: { [field: string]: Joi.Schema }
 }
--- a/packages/channels/src/messenger/channel.ts
+++ b/packages/channels/src/messenger/channel.ts
@@ -14,6 +14,7 @@ export class MessengerChannel extends ChannelTemplate<
     return {
       id: 'c4bb1487-b3bd-49b3-a3dd-36db908d165d',
       name: 'messenger',
+      version: '0.1.0',
       schema: MessengerConfigSchema,
       initiable: true,
       lazy: true
--- a/packages/server/src/channels/api.ts
+++ b/packages/server/src/channels/api.ts
@@ -8,16 +8,26 @@ export class ChannelApi {
 
   async setup() {
     const logger = this.app.logger.root.sub('channels')
-    const webhookRouter = Router()
+    const routers: { [version: string]: { router: Router; path: string } } = {
+      '0.1.0': {
+        router: Router(),
+        path: ''
+      },
+      '1.0.0': {
+        router: Router(),
+        path: '/v1'
+      }
+    }
 
     for (const channel of this.app.channels.list()) {
-      await channel.setup(webhookRouter, logger.sub(channel.meta.name))
+      const router = routers[channel.meta.version]
+      await channel.setup(router.router, logger.sub(channel.meta.name))
 
       channel.logger = this.app.logger.root.sub(channel.meta.name)
       channel.kvs = this.app.kvs
 
       channel.makeUrl(async (scope: string) => {
-        return `${process.env.EXTERNAL_URL}/webhooks/${scope}/${channel.meta.name}`
+        return `${process.env.EXTERNAL_URL}/webhooks${router.path}/${scope}/${channel.meta.name}`
       })
 
       channel.on('message', async ({ scope, endpoint, content }) => {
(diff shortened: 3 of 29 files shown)


packages/channels/src/base/meta.ts
```diff
@@ -3,7 +3,8 @@ import Joi from 'joi'
 export interface ChannelMeta {
   id: string
   name: string
+  version: string
   initiable: boolean
   lazy: boolean
-  schema: Joi.ObjectSchema
+  schema: { [field: string]: Joi.Schema }
 }
```
samuelmasse, line 10: Channels now have a version field. Notice they also provider the schema as an object now instead of a complete Joi.Schema. This is to allow us to add the version and enabled field to it.

packages/channels/src/messenger/channel.ts
```diff
@@ -14,6 +14,7 @@ export class MessengerChannel extends ChannelTemplate<
     return {
       id: 'c4bb1487-b3bd-49b3-a3dd-36db908d165d',
       name: 'messenger',
+      version: '0.1.0',
       schema: MessengerConfigSchema,
```
samuelmasse, line 18: Our legacy channels are retroactively considered to be version 0.1.0. The new channels will be version 1.0.0

packages/server/src/channels/api.ts
```diff
@@ -8,16 +8,26 @@ export class ChannelApi {
...
+        path: '/v1'
+      }
+    }
 
     for (const channel of this.app.channels.list()) {
-      await channel.setup(webhookRouter, logger.sub(channel.meta.name))
+      const router = routers[channel.meta.version]
+      await channel.setup(router.router, logger.sub(channel.meta.name))
 
       channel.logger = this.app.logger.root.sub(channel.meta.name)
```
samuelmasse, line 26: Channes with version 1.0.0 will have a /v1 in the webhook path while the old channels will use the path without /v1.

packages/server/src/channels/table.ts
```diff
@@ -8,8 +8,10 @@ export class ChannelTable extends Table {
 
   create(table: Knex.CreateTableBuilder) {
     table.uuid('id').primary()
-    table.string('name').unique().notNullable()
+    table.string('name').notNullable()
+    table.string('version').notNullable()
     table.boolean('lazy').notNullable()
     table.boolean('initiable').notNullable()
+    table.unique(['name', 'version'])
   }
```
samuelmasse, line 16: It's now possible to have multiple channels of the same name but of different versions. So there is a migration in this PR to handle that

packages/server/src/sync/api.ts
```diff
@@ -1,38 +1,65 @@
...
+                  ? Joi.string().valid(channel.meta.version).optional()
+                  : Joi.string().valid(channel.meta.version).required(),
+              enabled: Joi.boolean().optional(),
+              ...channel.meta.schema
+            }
+          }
+        }
+      })
+        .options({ stripUnknown: channel.meta.version === LEGACY_VERSION, abortEarly: false })
+        .validate({ body: { channels: { [channel.meta.name]: config } } })
```
samuelmasse, line 52: We only stripUknown for legacy channels. We will enforce correct payload formats for new channels. Here we're basically doing custom validation, but since we want the error messages to be consistent, we put the object to be validated inside body.channels so that all error message look like this : "error somewhere in body.channels.channelName etc"

packages/server/src/sync/schema.ts
```diff
@@ -2,21 +2,39 @@ import { Channel } from '@botpress/messaging-channels'
...
+        allowUnknown: true
+      })
+      .optional()
   }
 
   return ReqSchema({
     body: {
-      channels: Joi.object(channelsSchema).allow(null),
-      webhooks: Joi.array().items(SyncWebhookSchema).allow(null)
+      channels: Joi.object(channelsSchema).optional(),
```
samuelmasse, line 30: The first round of validation only checks if we are providing valid webhooks, as well as only existing channel names with existing versions

packages/server/src/migrations/1.0.2-channel-versions.ts
```diff
@@ -0,0 +1,60 @@
...
+    for (const newChannel of newChannels) {
+      await this.trx(getTableId('msg_channels'))
+        .update({ name: `${newChannel.name}@${newChannel.version}` })
+        .where({ id: newChannel.id })
+    }
+
+    await this.trx.schema.alterTable(getTableId('msg_channels'), (table) => {
+      table.dropUnique(['name', 'version'])
+      table.dropColumn('version')
+      table.string('name').unique().notNullable().alter()
```
laurentlp, line 57: Validate if this is equal: 

```suggestion
      table.unique(['name'])
```

laurentlp (approved):

PR #331, opened 2022-01-29
feat(slack): slack v1

samuelmasse:
Rewrite of slack channel. Went from using the old deprecated @slack/events-api and @slack/interactive-messages packages that were several minor and even major versions behind their latest version that is also deprecated. We now use the @slack/bolt package, which is a new official package. Also remove the two separate routes for webhooks into a single route

This is by far the most outdated channel I've updated yet

Closes DEV-2252


packages/channels/package.json
```diff
@@ -25,13 +25,15 @@
     "@types/uuid": "^8.3.4"
   },
   "dependencies": {
+    "@slack/bolt": "^3.9.0",
     "axios": "^0.24.0",
     "cli-color": "^2.0.1",
```
samuelmasse, line 30: That's the new packages that is recommended to use. It's an official slack package

packages/channels/src/messenger/README.md
```diff
@@ -1,29 +1,29 @@
...
-| Dropdown |   ✅   |
-| Card     |   ✅   |
-| Carousel |   ✅   |
-| File     |   ✅   |
-| Audio    |   ✅   |
-| Video    |   ✅   |
-| Location |   ✅   |
+| Channels | Messenger |
+| -------- | :-------: |
+| Text     |    ✅     |
```
samuelmasse, line 5: Unrelated. Forget to change the title here when messenger was merged.

packages/channels/src/slack/README.md
```diff
@@ -0,0 +1,29 @@
...
+| -------- | :---: |
+| Text     |  ✅   |
+| Image    |  ✅   |
+| Choice   |  ✅   |
+| Dropdown |  ✅   |
+| Card     |  ✅   |
+| Carousel |  ✅   |
+| File     |  ✅   |
+| Audio    |  ✅   |
+| Video    |  ✅   |
```
samuelmasse, line 13: These work by have markdown with a file url. There's a way to upload files to slack but I couldn't figure out how it work.

packages/channels/src/slack/api.ts
```diff
@@ -0,0 +1,170 @@
...
+    const signature = req.headers['x-slack-signature'] as string | undefined
+    const requestTimestamp = req.headers['x-slack-request-timestamp'] as string | undefined
+    const contentType = req.headers['content-type'] as string | undefined
+
+    if (!signature || !requestTimestamp) {
+      return res.sendStatus(401)
+    }
+
+    const ts = Number(requestTimestamp)
+    if (isNaN(ts)) {
```
samuelmasse, line 40: The code in this function is mostly copy paste from the code I found in https://github.com/slackapi/bolt-js/blob/main/src/receivers/ExpressReceiver.ts

I had to write a custom receiver that is adapted to being multi tenant.

packages/channels/src/slack/api.ts
```diff
@@ -0,0 +1,170 @@
...
+  private async handleRequest(req: ChannelApiRequest, res: Response) {
+    if (req.body?.ssl_check) {
+      return res.send()
+    } else if (req.body?.type === 'url_verification') {
+      return res.json({ challenge: req.body.challenge })
+    } else {
+      const event: ReceiverEvent = {
+        body: req.body,
+        ack: async (response) => {
+          if (!response) {
```
samuelmasse, line 85: Slack will validate the url before sending payloads, so we handle these validation steps here

packages/channels/src/slack/config.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export interface SlackConfig extends ChannelConfig {
+  botToken: string
+  signingSecret: string
+}
+
+export const SlackConfigSchema = {
+  botToken: Joi.string().required(),
+  signingSecret: Joi.string().required()
+}
```
samuelmasse, line 12: We might want to add more configs here as a future proof measure. For example ask the appId, etc. in case we need it in the future. For now these are the same configs as before

packages/channels/src/slack/service.ts
```diff
@@ -0,0 +1,23 @@
...
+      config,
+      app: new App({
+        signingSecret: config.signingSecret,
+        token: config.botToken,
+        tokenVerificationEnabled: false,
+        // We send a fake receiver here because we implement a receiver
+        // that works completely different than what bolt expects
+        receiver: { init: () => {} } as any
+      })
+    }
```
samuelmasse, line 21: See comment. It's kind of a hack but it works perfectly fine.

packages/channels/src/slack/README.md
laurentlp: Maybe you could use the ⚠️ emoji instead of the check, in this case?

packages/channels/src/slack/stream.ts
```diff
@@ -0,0 +1,26 @@
...
+  }
+
+  get senders(): ChannelSender<ChannelContext<any>>[] {
+    return [new TypingSender(), ...SlackSenders]
+  }
+
+  protected async getContext(base: ChannelContext<any>): Promise<SlackContext> {
+    return {
+      ...base,
+      message: { blocks: [] } as any
```
laurentlp, line 23: as any? Also, since we init blocks with an empty array. Which I think will always be the default, couldn't we change its typings so it's not optional anymore? This would allow us to remove a lot of null checking ('?').

packages/channels/src/slack/README.md
samuelmasse: Nah it's still useable. I would say this would be for stuff where you use actual functionality.

packages/server/test/integration/conversations.test.ts
```diff
@@ -72,7 +72,9 @@ describe('Conversations', () => {
   test('List conversations by user', async () => {
     const user = await users.create(state.client.id)
     const convo1 = await conversations.create(state.client.id, user.id)
+    await sleep()
```
laurentlp, line 75: Be careful by adding sleep in tests as this could lead to flaky tests. Also, the second it requires more than 10ms you'll need to up this value.

laurentlp (approved):

packages/channels/src/slack/api.ts
```diff
@@ -0,0 +1,170 @@
...
+      return res.sendStatus(401)
+    }
+
+    const ts = Number(requestTimestamp)
+    if (isNaN(ts)) {
+      return res.sendStatus(401)
+    }
+
+    // Divide current date to match Slack ts format
+    // Subtract 5 minutes from current time
```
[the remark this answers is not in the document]
samuelmasse: It tolerates payload that are at most 5 min old. This is copied from the slack client code

packages/channels/src/slack/api.ts
```diff
@@ -0,0 +1,170 @@
...
+  private async handleStart({ scope }: { scope: string }) {
+    const { app } = this.service.get(scope)
+    app.message(async (e) => this.handleMessage(scope, e))
+    app.action({}, async (e) => this.handleAction(scope, e))
+  }
+
+  private async handleMessage(scope: string, { message }: SlackEventMiddlewareArgs<'message'>) {
+    if ('bot_id' in message) {
+      return
+    }
```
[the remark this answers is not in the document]
samuelmasse: Yes

packages/server/test/integration/conversations.test.ts
samuelmasse: These tests are time bases though. Conversations are sorted by time created, but if they are inserted too fast the time created is the same, so these sleep fix that

packages/channels/src/slack/api.ts
```diff
@@ -0,0 +1,170 @@
...
+    const { config } = this.service.get(req.scope)
+    const stringBody = (await rawBody(req)).toString()
+
+    const hmac = crypto.createHmac('sha256', config.signingSecret)
+    const [version, hash] = signature.split('=')
+    hmac.update(`${version}:${ts}:${stringBody}`)
+
+    if (!tsscmp(hash, hmac.digest('hex'))) {
+      return res.sendStatus(403)
+    }
```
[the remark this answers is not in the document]
samuelmasse: Added a comment

packages/channels/src/slack/stream.ts
samuelmasse: Good point. Removed the as any and added typings to avoid null checking

packages/server/test/integration/conversations.test.ts
laurentlp: Ahhh I see my bad!!

PR #336, opened 2022-02-02
feat(teams): teams v1

samuelmasse:
Upgrades teams to be version 1.0.0. Added more content types and fixed some things. As for the API version nothing really changed for this channel. We were using botbuilder 4.14.0 before and now we use 4.15.0

Closes DEV-2277


packages/channels/package.json
```diff
@@ -27,6 +27,7 @@
   "dependencies": {
     "@slack/bolt": "^3.9.0",
     "axios": "^0.25.0",
+    "botbuilder": "^4.15.0",
     "cli-color": "^2.0.1",
```
samuelmasse, line 31: We've upgraded a whopping 1 minor version here (from 4.14.0). I think I may have updated this channel before.

packages/channels/src/teams/README.md
```diff
@@ -0,0 +1,29 @@
...
+| -------- | :---: |
+| Text     |  ✅   |
+| Image    |  ✅   |
+| Choice   |  ✅   |
+| Dropdown |  ✅   |
+| Card     |  ✅   |
+| Carousel |  ✅   |
+| File     |  ✅   |
+| Audio    |  ✅   |
+| Video    |  ✅   |
```
samuelmasse, line 13: We send these as text with a link

packages/channels/src/teams/README.md
```diff
@@ -0,0 +1,29 @@
...
+| Location |  ✅   |
+
+### Receiving
+
+| Channels      | Teams |
+| ------------- | :---: |
+| Text          |  ✅   |
+| Quick Reply   |  ✅   |
+| Postback      |  ✅   |
+| Say Something |  ✅   |
```
samuelmasse, line 23: I sorted these out. Turns out this channel only received text, even if it was quick reply, postback and say something. Now all three work.

packages/channels/src/teams/config.ts
```diff
@@ -0,0 +1,12 @@
+import Joi from 'joi'
+import { ChannelConfig } from '../base/config'
+
+export interface TeamsConfig extends ChannelConfig {
+  appId: string
+  appPassword: string
+}
```
samuelmasse, line 7: Removed the tenantId here because I don't know what it's used for. We can add it back if we need it.

laurentlp: I think it was for some validation but I'm not sure either.

laurentlp (approved):

PR #338, opened 2022-02-03
feat(vonage): vonage v1

samuelmasse:
Upgrades vonage to use the messages api 1.0 instead of 0.1. There is no longer two webhooks and both inbound and status come in at /vonage. The amount of configuration required as also been severely reduced (only apiKey, apiSecret, and signatureSecret).

Added interactive messages for quick replies

TODO: 
- add reception for all content types. 
- possibly use interactive messages for carousels
- add typings

Closes DEV-2278


packages/client/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@botpress/messaging-client",
-  "version": "1.0.0",
+  "version": "1.0.1",
```
samuelmasse, line 3: Unrelated. This version is already released but I didn't commit the version change for some reason so this fixes it.

samuelmasse:
Note: video seems to not work despite sending the correct payload

samuelmasse:
Not sure what happened to the actions here

packages/channels/package.json
```diff
@@ -31,6 +31,7 @@
     "cli-color": "^2.0.1",
     "express": "^4.17.2",
     "joi": "^17.6.0",
```
samuelmasse, line 33: I tried using the @vonage/server-sdk packages but it's a total mess. You can only make sms request with it, for some reason they dropped the whatsapp functions between version 2.10.7 and 2.10.10. So instead we just make the http request ourselves

packages/channels/src/vonage/README.md
```diff
@@ -0,0 +1,29 @@
...
+| Text          |   ✅   |
+| Quick Reply   |   ✅   |
+| Postback      |   ✅   |
+| Say Something |   ✅   |
+| Voice         |   ❌   |
+| Image         |   ❌   |
+| File          |   ❌   |
+| Audio         |   ❌   |
+| Video         |   ❌   |
+| Location      |   ❌   |
```
samuelmasse, line 29: Need to add this since the previous version had it

packages/channels/src/vonage/api.ts
```diff
@@ -0,0 +1,73 @@
...
+
+    if (req.body.reply) {
+      const [_, payload] = req.body.reply.id.split('::')
+
+      await this.service.receive(req.scope, endpoint, {
+        type: 'quick_reply',
+        text: req.body.reply.title,
+        payload
+      })
+    } else {
```
samuelmasse, line 59: We have quick replies now due to the new interactive messages feature https://developers.facebook.com/docs/whatsapp/guides/interactive-messages

packages/channels/src/vonage/config.ts
```diff
@@ -0,0 +1,16 @@
+import Joi from 'joi'
+import { ChannelConfig } from '../base/config'
+
+export interface VonageConfig extends ChannelConfig {
+  apiKey: string
+  apiSecret: string
+  signatureSecret: string
+  useTestingApi?: boolean
```
samuelmasse, line 8: I didn't find a reason yet to require the applicationId and the privateKey here so I removed them

samuelmasse:
We can use this to test vonage templates :
```ts
  {
    "type": "vonage-template",
    "namespace": "9b6b4fcb_da19_4a26_8fe8_78074a91b584",
    "name": "verify",
    "parameters": ["yo", "bla", "cra"]
  },
  {
    "type": "vonage-media-template",
    "namespace": "9b6b4fcb_da19_4a26_8fe8_78074a91b584",
    "name": "sandbox_travel_boardingpass",
    "header": {
      "parameters": [
        {
          "type": "image",
          "image": {
            "link": "https://upload.wikimedia.org/wikipedia/commons/a/a1/Mallard2.jpg"
          }
        }
      ]
    },
    "body": {
      "parameters": [
        {
          "type": "text",
          "text": "Value 1"
        },
        {
          "type": "text",
          "text": "Value 2"
        },
        {
          "type": "text",
          "text": "Value 3"
        },
        {
          "type": "text",
          "text": "Value 4"
        }
      ]
    }
  }
```

samuelmasse:
Note: video is sent correctly but is not displayed

laurentlp (approved):

PR #341, opened 2022-02-07
chore(webchat): fixes and planning

samuelmasse:
Closes DEV-2289

### DIFF
--- a/packages/webchat/src/main.tsx
+++ b/packages/webchat/src/main.tsx
@@ -17,7 +17,7 @@ import { Config, Message, Overrides, uuid } from './typings'
 import { checkLocationOrigin, initializeAnalytics, isIE, trackMessage, trackWebchatState } from './utils'
 
 const _values = (obj: Overrides) => Object.keys(obj).map((x) => obj[x])
-const DEFAULT_TYPING_DELAY = 500
+const DEFAULT_TYPING_DELAY = 1000
 
 class Web extends React.Component<MainProps> {
   private config!: Config


samuelmasse:
Simply refreshing the webchat produces empty messages that shouldn't be displayed

Screenshot from 2022-02-07 16-29-35

samuelmasse:
The single-choice message does not display a keyboard with buttons to click

Screenshot from 2022-02-07 16-35-24

Should look like this

Screenshot from 2022-02-07 16-35-15

samuelmasse:
Cards are not displayed correctly (problem with width)

Screenshot from 2022-02-07 17-13-13

Should look like this

Screenshot from 2022-02-07 17-13-03

samuelmasse:
Images are way too large

Screenshot from 2022-02-07 17-16-35

Should be like this (with title added)

Screenshot from 2022-02-07 17-17-12

samuelmasse:
Audio should have new line separating the title

Screenshot from 2022-02-07 17-33-14

samuelmasse:
There's probably a prettier way to display this title

Screenshot from 2022-02-07 17-35-20

samuelmasse:
Location content is not supported in either implementation. We could add it pretty quickly. The only thing that is needed is a google maps link that we generate

samuelmasse:
This one I'm not sure about. We don't use the keyboard anymore to display the options. Everything seems to work ok. The new version seems better on desktop, but the old one would be better on mobile

Screenshot from 2022-02-07 17-48-21

Before

Screenshot from 2022-02-07 17-48-28

samuelmasse:
Elements that were imported from ui-shared-lite such as the Tooltip are commented out at a few places. They should be brought back in the webchat code to be used.

samuelmasse:
The sendEvent function from the api is commented out. It used to send events directly. Something we could do to replace it is add a "transient" property the messages route. Something else would be to add a conversation.visit event to messaging, which is what the sendEvent function was used for.

samuelmasse:
Fetch preferences doesn't exist anymore. Right now we hardcode that the user's language is english

samuelmasse:
Sending voice messages doesn't work

samuelmasse:
Conversation references don't work (not even sure what they are used for)

samuelmasse:
Can't set a custom user id

laurentlp (approved):

PR #351, opened 2022-02-10
chore(api): add /v1 to api paths

samuelmasse:
Adds /v1 to API paths. Is backward compatible with not using /v1. Will require a new release of the http client.

Also should we version the socket connection? Right now the socket connects at the root of our url. Maybe it could connect at `/sockets/v1`?

The client package needs to be updated

Closes DEV-2285
Closes DEV-2323

### DIFF
--- a/packages/client/package.json
+++ b/packages/client/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@botpress/messaging-client",
-  "version": "1.0.2",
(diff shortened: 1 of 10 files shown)


laurentlp (approved):
LGTM! 

Maybe we should consider adding the v1 path to the socket API only when we'll really need it?

PR #357, opened 2022-02-14
feat(webchat): builtin visit event

samuelmasse:
Adds a conversation.visit event that replaces the old user_visit message. We'll need to add handling for this on botpress

TODO: need to add handling to receive this event in the http client

Closes DEV-2309

### DIFF
--- a/packages/webchat/src/store/index.ts
+++ b/packages/webchat/src/store/index.ts
@@ -229,7 +229,6 @@ class RootStore {
     }
 
     await this.fetchPreferences()
-    await this.sendUserVisit()
   }
 
   @action.bound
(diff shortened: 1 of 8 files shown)


packages/webchat/src/store/index.ts
```diff
@@ -353,12 +353,10 @@ class RootStore {
...
-    await this.sendData({
-      type: 'visit',
-      text: 'User visit',
-      timezone: -(new Date().getTimezoneOffset() / 60), // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getTimezoneOffset#description
-      language: getUserLocale()
-    })
+    await this.api.sendVisit(
+      -(new Date().getTimezoneOffset() / 60), // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getTimezoneOffset#description
+      getUserLocale()
+    )
```
samuelmasse, line 359: The original message had the timezone and language included in the event. Maybe we should not include these in the event yet, since we don't really handle anything language related at the moment? It would make more sense to include these in user.new, or conversation.new, or conversation.started. It would be odd to have the locale or timezone change after a visit (could happen though). Also do we even use the timezone?

samuelmasse: I suggest we remove this for now, as to not commit to a feature that might be badly placed. We can add this back later if we need it. From what I could find in the codebase, the visit event is only used by the channel web to add the timezone and locale as user attributes

laurentlp: I agree with you here. Let's remove them and add them back if needed later on!

PR #360, opened 2022-02-14
feat(mapping): endpoint api

samuelmasse:
Adds two new routes that make up the "endpoint" API. One route `/endpoints/map` to get a conversationId from an endpoint, and the route `/endpoints/conversation/:conversationId` to list the endpoints associated to a conversation (since a conversation can be present on more than one channel)

TODO: add documentation

Closes DEV-2339

### DIFF
--- a/packages/client/test/e2e/client.test.ts
+++ b/packages/client/test/e2e/client.test.ts
@@ -356,9 +356,29 @@ describe('Http Client', () => {
         expect(convoId2).toBe(convoId)
       })
 
-      test('Should be able to reverse map the conversation id to get back the same endpoint', async () => {
+      test('Should fail to map an endpoint with an unknown channel', async () => {
+        await expect(client.mapEndpoint({ ...endpoint, channel: { name: 'yoyo', version: '1.0.0' } })).rejects.toThrow(
+          new Error('Request failed with status code 400')
+        )
+      })
+
+      test('Should fail to map an endpoint with invalid fields', async () => {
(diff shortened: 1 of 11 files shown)


packages/server/src/mapping/api.ts
```diff
@@ -0,0 +1,52 @@
...
+  async map(req: ClientApiRequest, res: Response) {
+    const endpoint = req.body as Endpoint
+
+    const channel = this.channels.getByNameAndVersion(endpoint.channel.name, endpoint.channel.version)
+    const { conversationId } = await this.mapping.getMapping(req.clientId, channel.meta.id, endpoint)
+
+    res.send({ conversationId })
+  }
+
+  async revmap(req: ClientApiRequest, res: Response) {
```
laurentlp:
```suggestion
  async listMap(req: ClientApiRequest, res: Response) {
```

packages/client/test/e2e/client.test.ts
```diff
@@ -340,4 +340,26 @@ describe('Http Client', () => {
       })
     })
   })
+
+  describe('Endpoints', () => {
+    describe('Map', () => {
+      const endpoint = { channel: { name: 'telegram', version: '1.0.0' }, identity: '*', sender: 'yoyo', thread: 'ya' }
+      let convoId: string | undefined
+
```
laurentlp, line 348: Add those scenarios:

- Calling listEndpoints when no mapping was created should return a 404 or empty array.
- Try creating an endpoint mapping with an invalid channel.
- Try creating an endpoint mapping with empty identity, sender, and thread.

laurentlp (approved):

PR #363, opened 2022-02-16
feat(webchat): send conversation.started event

samuelmasse:
Webchat now sends a conversation.started even when it creates a conversation.

Closes DEV-2309


laurentlp (approved):
LGTM!

PR #365, opened 2022-02-16
feat(channels): disable legacy channels by default

samuelmasse:
Makes it so legacy channels are disabled by default, and need to be enabled with ENABLE_LEGACY_CHANNELS

Closes DEV-2353


laurentlp (approved):
LGTM

PR #367, opened 2022-02-16
chore(webchat): fix display of single-choice

laurentlp:
This PR fixes the rendering of single-choices.

**Screenshot**

https://user-images.githubusercontent.com/9640576/154366205-d0ad56b0-293b-4cf9-819e-fcc4d2676f6e.mp4

Closes DEV-2294


packages/components/src/renderer/keyboard.tsx
```diff
@@ -26,9 +26,9 @@ class KeyboardElements extends React.Component<KeyboardElementsProps> {
   }
 
   componentWillUnmount() {
-    if (this.props.append) {
+    if (this.props.append && Keyboard.appendRef.current?.hasChildNodes()) {
       Keyboard.appendRef.current?.removeChild(this.container)
-    } else {
+    } else if (!this.props.append && Keyboard.prependRef.current?.hasChildNodes()) {
```
laurentlp, line 31: The `"**Cannot display message**"` error was caused by the fact that we tried to remove the container from the ref's children when it wasn't one.

packages/components/src/renderer/choice.tsx
```diff
@@ -20,7 +20,7 @@ export class SingleChoice extends Component<MessageTypeHandlerProps<'single-choi
   }
 
   handleButtonClicked = (title: string, payload: any) => {
-    this.props.config.onSendData?.({
+    void this.props.config.onSendData?.({
```
laurentlp, line 23: Fixes a linting issue

samuelmasse (approved):
NICE

PR #378, opened 2022-02-18
chore(webchat): removed emulator related code

laurentlp:
This PR aims at removing all code related to the emulator.

Closes DEV-2307

### DIFF
--- a/packages/webchat/src/main.tsx
+++ b/packages/webchat/src/main.tsx
@@ -24,7 +24,6 @@ class Web extends React.Component<MainProps> {
   private parentClass!: string
   private hasBeenInitialized: boolean = false
   private audio!: HTMLAudioElement
-  private lastMessageId!: uuid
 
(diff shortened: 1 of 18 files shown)


packages/webchat/src/main.tsx
```diff
@@ -103,9 +91,10 @@ class Web extends React.Component<MainProps> {
       this.postMessageToParent('setWidth', this.config.containerWidth)
     }
 
-    if (this.config.reference) {
-      await this.props.setReference!()
-    }
+    // TODO: properly implement of deprecate this feature
+    //if (this.config.reference) {
+    //  await this.props.setReference!()
+    //}
```
samuelmasse, line 97: I'm pretty sure this feature just stops existing

laurentlp: Well, I think so too but since I'm not 100% sure, I preferred simply commenting it out.

PR #380, opened 2022-02-18
chore(channels): add version and down rendering info to channels readme

laurentlp:
This PR adds versions to the channels README and information about which content-element is down rendered on which channel.

### DIFF
--- a/packages/channels/src/messenger/README.md
+++ b/packages/channels/src/messenger/README.md
@@ -2,18 +2,18 @@
 
 ### Sending
 
-| Channels | Messenger | Details               |
-| -------- | :-------: | :-------------------- |
-| Text     |    ✅     |                       |
-| Image    |    ✅     |                       |
-| Choice   |    ✅     | Choices down rendered |
-| Dropdown |    ✅     | Choices down rendered |
-| Card     |    ✅     |                       |
-| Carousel |    ✅     |                       |
-| File     |    ✅     | File sent as text     |
-| Audio    |    ✅     | Audio sent as text    |
-| Video    |    ✅     | Video sent as text    |
-| Location |    ✅     | Location sent as text |
+| Channels | Messenger | Details                    |
+| -------- | :-------: | :------------------------- |
+| Text     |    ✅     |                            |
+| Image    |    ✅     |                            |
+| Choice   |    ✅     | Choices down rendered      |
+| Dropdown |    ✅     | Choices down rendered      |
+| Card     |    ✅     |                            |
+| Carousel |    ✅     |                            |
+| File     |    ✅     | File sent as text link     |
+| Audio    |    ✅     | Audio sent as text link    |
+| Video    |    ✅     | Video sent as text link    |
+| Location |    ✅     | Location sent as text link |
 
 ### Receiving
 
--- a/packages/channels/src/messenger/README.md
+++ b/packages/channels/src/messenger/README.md
@@ -1,3 +1,5 @@
+# Messenger (v1.0.0)
+
 ### Sending
 
 | Channels | Messenger |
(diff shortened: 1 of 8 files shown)


samuelmasse:
This looks like a draft?

packages/channels/src/messenger/README.md
```diff
@@ -1,29 +1,31 @@
...
+| Text     |    ✅     |                            |
+| Image    |    ✅     |                            |
+| Choice   |    ✅     |                            |
+| Dropdown |    ✅     |                            |
+| Card     |    ✅     |                            |
+| Carousel |    ✅     |                            |
+| File     |    ✅     | File sent as text link     |
+| Audio    |    ✅     | Audio sent as text link    |
+| Video    |    ✅     | Video sent as text link    |
+| Location |    ✅     | Location sent as text link |
```
samuelmasse, line 16: We should say it's a google maps link here and for all the other channels that down render locations the same way

packages/channels/src/messenger/README.md
```diff
@@ -1,29 +1,31 @@
...
-| Location |    ✅     |
+| Channels | Messenger | Details                    |
+| -------- | :-------: | :------------------------- |
+| Text     |    ✅     |                            |
+| Image    |    ✅     |                            |
+| Choice   |    ✅     |                            |
+| Dropdown |    ✅     |                            |
+| Card     |    ✅     |                            |
+| Carousel |    ✅     |                            |
+| File     |    ✅     | File sent as text link     |
```
samuelmasse, line 13: Why not `sent as url`? The more wordy and correct way to say it would be `text for title with a link to the file`. If we want to be more concise it's fine but why use `text link` as a sort of roundabout way of saying that there's a text part and a link part.

samuelmasse (approved):

packages/channels/src/twilio/README.md
```diff
@@ -1,29 +1,31 @@
...
-| Location |   ✅   |
+| Channels | Twilio | Details                    |
+| -------- | :----: | :------------------------- |
+| Text     |   ✅   |                            |
+| Image    |   ✅   |                            |
+| Choice   |   ✅   | Choices down rendered      |
+| Dropdown |   ✅   | Choices down rendered      |
+| Card     |   ✅   | Buttons down rendered      |
+| Carousel |   ✅   |                            |
+| File     |   ✅   |                            |
```
samuelmasse, line 13: Files aren't sent as links for twilio?

packages/channels/src/vonage/README.md
```diff
@@ -1,29 +1,31 @@
...
-| Video    |   ✅   |
-| Location |   ✅   |
+| Channels | Vonage | Details               |
+| -------- | :----: | :-------------------- |
+| Text     |   ✅   |                       |
+| Image    |   ✅   |                       |
+| Choice   |   ✅   |                       |
+| Dropdown |   ✅   |                       |
+| Card     |   ✅   | Buttons down rendered |
+| Carousel |   ✅   |                       |
```
samuelmasse, line 12: This applies to both card and carousel

packages/channels/src/twilio/README.md
laurentlp: Right, good catch!

packages/channels/src/twilio/README.md
```diff
@@ -1,29 +1,31 @@
...
-| Video    |   ✅   |
-| Location |   ✅   |
+| Channels | Twilio | Details                           |
+| -------- | :----: | :-------------------------------- |
+| Text     |   ✅   |                                   |
+| Image    |   ✅   |                                   |
+| Choice   |   ✅   | Choices down rendered             |
+| Dropdown |   ✅   | Choices down rendered             |
+| Card     |   ✅   | Buttons down rendered             |
+| Carousel |   ✅   |                                   |
```
samuelmasse, line 12: Applies to both here as well

PR #384, opened 2022-02-22
chore(webchat): store and load user lang in browser storage

laurentlp:
This PR adds the storage class (copied from ui-lite) and uses it to store and fetch the user language in the browser's storage (local, session, cookie).


packages/webchat/src/core/api.tsx
```diff
@@ -13,12 +13,14 @@ export default class WebchatApi {
 
   constructor(private socket: BpSocket) {}
 
+  // TODO: Do we still need this?
   private get baseUserPayload() {
     return {
       webSessionId: window.__BP_VISITOR_SOCKET_ID
     }
   }
```
samuelmasse, line 21: We don't need this no we can delete it.

packages/webchat/src/core/api.tsx
```diff
@@ -13,12 +13,14 @@ export default class WebchatApi {
...
 
+  // TODO: Fix this
   async fetchBotInfo() {
     try {
       const { data } = await this.axios.get('/botInfo', this.axiosConfig)
```
samuelmasse, line 26: This one will be done by fetching the botInfo.json file that will be located at https://mediafiles.botpress.dev/clientId/info.json

packages/webchat/src/core/api.tsx
```diff
@@ -78,6 +63,7 @@ export default class WebchatApi {
     }
   }
 
+  // TODO: Fis this
   async resetSession(conversationId: uuid): Promise<void> {
     try {
       await this.axios.post('/conversations/reset', { ...this.baseUserPayload, conversationId }, this.axiosConfig)
```
samuelmasse, line 69: Maybe this should be considered debugger stuff? If a user wants a new conversation we can simply make an API call to create a new conversation. There is no need for a "reset conversation" concept

packages/webchat/src/core/api.tsx
```diff
@@ -103,6 +89,7 @@ export default class WebchatApi {
     }
   }
 
+  // TODO: Fix this
   async downloadConversation(conversationId: uuid): Promise<any> {
     try {
       const { data } = await this.axios.post(
```
samuelmasse, line 95: That's doable in frontend

packages/webchat/src/core/api.tsx
```diff
@@ -135,6 +122,7 @@ export default class WebchatApi {
     }
   }
 
+  // TODO: Fix this
   async deleteMessages(conversationId: uuid) {
     try {
       await this.axios.post(
```
samuelmasse, line 128: There's a API route already to delete all messages of a conversation. We could add a way to do this by socket as well.

packages/webchat/src/utils/storage.ts
```diff
@@ -0,0 +1,103 @@
...
+let keyPrefix = ''
+const setKeyPrefix = (prefix: string) => {
+  keyPrefix = prefix
+}
+
+const formatKey = (key: string) => {
+  return `${keyPrefix}-${key}`
+}
+
+let useSessionStorage = new Boolean(window.USE_SESSION_STORAGE)
```
samuelmasse, line 19: When is USE_SESSION_STORAGE true?

packages/webchat/src/store/composer.ts
```diff
@@ -52,7 +50,7 @@ class ComposerStore {
       this._sentHistoryIndex = 0
 
       if (this.rootStore.config.enablePersistHistory) {
-        window.BP_STORAGE?.set(SENT_HISTORY_KEY, takeRight(this._sentHistory, constants.SENT_HISTORY_SIZE))
+        window.BP_STORAGE.set(SENT_HISTORY_KEY, takeRight(this._sentHistory, constants.SENT_HISTORY_SIZE))
```
samuelmasse, line 53: BP_STORAGE is defined? Where is that created?

samuelmasse (commented):
How do we test this? Also it seems like the data about language is stored in a different way than the user creds. They should probably use the same code. I'm also not a fan of having a global window.BP_STORAGE item, just importing functions would be better.

packages/webchat/src/core/api.tsx
laurentlp: Right! Not sure if this was the case, but maybe this code should have been removed when I deleted the emulator code?

packages/webchat/src/core/api.tsx
laurentlp: Perfect adding to the QA task.

packages/webchat/src/store/composer.ts
laurentlp: It's kind of a hack, but I import it in the root index.tsx file. I think we should rethink the storage class and make something cleaner and simpler. Maybe in another PR tho?

samuelmasse (approved):

PR #387, opened 2022-02-22
chore(webchat): various improvements

laurentlp:
What was done:

 - [X] When no image is shown, we should display an alternative (alt) text different than the file title as we already display it above the image
    - We don't need to display an alt text as the title will always be shown. We cannot basically provide more info than what we already display. It then makes more sense not to display the title twice by also having it in the alt attribute.
- [X] Pressing enter on android doesn't send the message. Enter needs to be pressed 2 times
    - Cannot reproduce.
 - [X] If you refresh during a single choice the choices disappear. This didn't happen with the old one 
    - The issue was that the keyboard was rendered after the single choice. Meaning that choices could not be prepended to the non-existing keyboard. We now wait for the keyboard to be rendered before rendering the MessageList (kind of hackish but it works). I could not find a better solution.
 - [X] Clean up dead code: https://github.com/botpress/messaging/pull/384#discussion_r811501560
 - [X] Download conversation in front-end: https://github.com/botpress/messaging/pull/384#discussion_r811502504
 - [X] Add path to delete conversations with sockets: https://github.com/botpress/messaging/pull/384#discussion_r811503045
 - [X] Fetch bot info https://github.com/botpress/messaging/pull/384#discussion_r811502017
 - [X] Reset session could just create a new conversation: https://github.com/botpress/messaging/pull/384#discussion_r811502434
    - This issue needs clarification as this behavior is similar to the delete conversation feature (_skipping for now_).
 - [X] Refactor storage class. See this PR: https://github.com/botpress/messaging/pull/384

Config used to test all features:

```jsonc
{
  messagingUrl: 'http://localhost:3100',
  clientId: '',
  encryptionKey: '1234567890',
  enableReset: false, // We don't know yet what to do with this
  enableConversationDeletion: true,
  showConversationsButton: true,
  useSessionStorage: false // Toggle this value to store the creds in session or local storage
}
```

Closes DEV-2360


samuelmasse (approved):
Awesome work! Tested and working :+1:

PR #398, opened 2022-03-02
chore(tests): added possibility to seed the database when running tests

laurentlp:
This PR adds a class that can seed data to some of the tables we interact with on our migrations.


samuelmasse:
We should move the utils folder out from the integration folder since it's used by migration tests also

samuelmasse (approved):

PR #399, opened 2022-03-02
chore(server): use framework

samuelmasse:
Adds the framework from the other repo, changes messaging to use the framework and adds a basic example framework project

Todo: 
- Add back clients/sync route
- Write migration for provision table
- Write migration to transfer msg_clients.providerId -> provisions table
- Stop copying client id to the provider name column. Provider name is an override, otherwise use the client id pointed by the provision
- Write migration to remove the provider names that are client ids

Closes DEV-2388


samuelmasse:
This PR was too large so split up in https://github.com/botpress/messaging/pull/411 and https://github.com/botpress/messaging/pull/407

samuelmasse:
This PR was too large so split up in https://github.com/botpress/messaging/pull/411 and https://github.com/botpress/messaging/pull/407

PR #400, opened 2022-03-03
chore(client): cookie handling for load balancer

samuelmasse:
Adds handling in the messaging client to use the bp_messaging cookie for load balancing

Closes DEV-2400

### DIFF
--- a/packages/client/src/base.ts
+++ b/packages/client/src/base.ts
@@ -105,7 +105,7 @@ export abstract class MessagingChannelBase extends Emitter<{
     for (const strCookie of cookieHeader) {
       const resCookie = cookie.parse(strCookie)
       if (resCookie[this.sessionCookieName]) {
-        this.http.defaults.headers.common['cookie'] = `${this.sessionCookieName}=${resCookie.bp_messaging};`
+        this.http.defaults.headers.common['cookie'] = `${this.sessionCookieName}=${resCookie[this.sessionCookieName]};`
       }
     }
   }
--- a/packages/client/src/base.ts
+++ b/packages/client/src/base.ts
@@ -1,5 +1,6 @@
 import { uuid } from '@botpress/messaging-base'
 import axios, { AxiosInstance, AxiosRequestConfig } from 'axios'
+import cookie from 'cookie'
 import { ConversationStartedEvent, MessageFeedbackEvent, MessageNewEvent, UserNewEvent } from '.'
 import { MessagingClientAuth } from './auth'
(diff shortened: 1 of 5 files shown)


samuelmasse:
Note: we'll need to release a new version of the messaging client and update it on the runtime to see the change in behavior

packages/client/src/base.ts
```diff
@@ -72,8 +73,32 @@ export abstract class MessagingChannelBase extends Emitter<{
...
+  }
+
+  private saveCookie(cookieHeader: string[] | undefined) {
+    if (!cookieHeader) {
+      return
+    }
+
+    for (const strCookie of cookieHeader) {
+      const resCookie = cookie.parse(strCookie)
+      if (resCookie.bp_messaging) {
```
michaelmass, line 98: The name of the `cookie` should be extracted in a configuration setting this way it is going to be easier to change it without having to rebuild a new image of the runtime.

```
const config = {
   [...]
   messagingSessionCookie: process.env.MESSAGING_SESSION_COOKIE ?? 'bp_messaging'
}
```

michaelmass (changes requested):

packages/client/src/base.ts
```diff
@@ -72,8 +82,32 @@ export abstract class MessagingChannelBase extends Emitter<{
...
+
+  private saveCookie(cookieHeader: string[] | undefined) {
+    if (!this.sessionCookieName || !cookieHeader) {
+      return
+    }
+
+    for (const strCookie of cookieHeader) {
+      const resCookie = cookie.parse(strCookie)
+      if (resCookie[this.sessionCookieName]) {
+        this.http.defaults.headers.common['cookie'] = `${this.sessionCookieName}=${resCookie.bp_messaging};`
```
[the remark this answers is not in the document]
samuelmasse: Yup good catch

packages/client/src/options.ts
```diff
@@ -12,6 +12,8 @@ export interface MessagingChannelOptions {
   axios?: Omit<AxiosRequestConfig, 'baseURL'>
   /** Optional logger interface that can be used to get better debugging */
   logger?: Logger
+  /** Name of the cookie for sticky sessions */
+  sessionCookieName?: string
 }
 
 export interface MessagingOptions extends Omit<MessagingChannelOptions, 'adminKey'>, MessagingClientAuth {
```
samuelmasse, line 19: @michaelmass Added the sessionCookieName config

michaelmass (approved):
LGTM

PR #401, opened 2022-03-04
chore(tests): added global diff migration test

laurentlp:
This PR adds the global diff migration test to sure that the code describing the tables is synced correctly with the migrations.

### DIFF
--- a/packages/server/test/migration/global-diff.test.ts
+++ b/packages/server/test/migration/global-diff.test.ts
@@ -1,0 +1,77 @@
+import schemaInspector from 'knex-schema-inspector'
+import { setupConnection } from './utils/database'
+
+import { startMessagingServer } from './utils/server'
+
+const NO_MIGRATION = 'no_mig'
+const ALL_MIGRATIONS = 'all_migs'
+const TIMEOUT = 30000
+
+describe('Global Diff', () => {
+  test(
+    'Starts Messaging with the latest database schema',
+    async () => {
+      await startMessagingServer(
+        {
+          command: 'yarn dev',
+          launchTimeout: TIMEOUT,
+          protocol: 'http',
+          host: '127.0.0.1',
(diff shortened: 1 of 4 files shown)


packages/server/test/migration/global-diff.test.ts
```diff
@@ -0,0 +1,77 @@
...
+        )
+
+        for (const { table, column } of noMigrationTableColumns) {
+          const noMigrationTableColumnInfo = await noMigrationInspector.columnInfo(table, column)
+          const allMigrationsTableColumnInfo = await allMigrationsInspector.columnInfo(table, column)
+
+          expect(noMigrationTableColumnInfo).toEqual(allMigrationsTableColumnInfo)
+        }
+
+        expect(noMigrationTableColumns.length).toEqual(allMigrationsTableColumns.length)
```
laurentlp, line 68: I have to compare columns by columns for each table since the SQL used to create the tables differs slightly when ran from migrations than the code.

packages/server/test/migration/utils/database.ts
```diff
@@ -0,0 +1,73 @@
...
+import { SchemaInspector } from 'knex-schema-inspector/dist/types/schema-inspector'
+import { Table } from 'knex-schema-inspector/dist/types/table'
+import path from 'path'
+
+export class Inspector {
+  private _knex: Knex
+  private _url: string
+  private _isLite: boolean = false
+  private _inspector: SchemaInspector
+
```
samuelmasse, line 12: Why underscores? We don't usually use them in the codebase

laurentlp: Personal preference. Let's add an ESLint rule to enforce it later on. Something like: https://eslint.org/docs/rules/no-underscore-dangle

samuelmasse (approved):

PR #402, opened 2022-03-09
fix(sockets): fix socket unable to send error message crashing server

samuelmasse:
Fixes bug reported here https://botpresshq.slack.com/archives/C02AFENA9C5/p1646837902625429

The code for socket message handling had a try catch, but the code located in the catch could also throw, so I added a try catch there also.

Closes DEV-2425

### DIFF
--- a/packages/server/src/socket/manager.ts
+++ b/packages/server/src/socket/manager.ts
@@ -136,11 +136,11 @@ export class SocketManager {
   }
 
   async setupSocket(socket: Socket.Socket) {
-    socket.on('message', async (data) => {
-      await this.handleSocketMessage(socket, data)
+    socket.on('message', (data) => {
+      void this.handleSocketMessage(socket, data)
     })
-    socket.on('disconnect', async () => {
-      await this.handleSocketDisconnect(socket)
+    socket.on('disconnect', () => {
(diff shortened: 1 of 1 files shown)


packages/server/src/socket/manager.ts
```diff
@@ -136,11 +136,11 @@ export class SocketManager {
...
   async setupSocket(socket: Socket.Socket) {
-    socket.on('message', async (data) => {
-      await this.handleSocketMessage(socket, data)
+    socket.on('message', (data) => {
+      void this.handleSocketMessage(socket, data)
     })
-    socket.on('disconnect', async () => {
-      await this.handleSocketDisconnect(socket)
+    socket.on('disconnect', () => {
+      void this.handleSocketDisconnect(socket)
```
laurentlp, line 143: Why void?

laurentlp (approved):
Small question otherwise LGTM

packages/server/src/socket/manager.ts
samuelmasse: The `on` handler isn't async, so I put void here to indicate cleary that these are floating promises, and that the code needs to be correctly try catched otherwise it crashes the whole server

PR #407, opened 2022-03-10
chore(framework): bring framework code

samuelmasse:
Brings the framework code from the other repo to messaging. Also adds an example project that uses it. Makes no changes to any other packages

Closes DEV-2388


samuelmasse:
Something weird happening with docker. yarn --immutable expects everything to be there, but we might not want to do that with the docker image

PR #409, opened 2022-03-10
feat(components): added location component

laurentlp:
This PR adds the location component and enables its support on the webchat.

It requires a Google Maps API Key to be set in the webchat config: https://developers.google.com/maps/documentation/javascript/get-api-key

Screenshot from 2022-03-10 17-47-44

**api.rest** 

```
### Send Location
POST {{baseUrl}}/api/v1/messages
Authorization: Bearer {{authToken}}
x-bp-messaging-client-id: {{clientId}}
x-bp-messaging-client-token: {{clientToken}}
Content-Type: application/json

{
  "conversationId": "{{convoId}}",
  "payload": {
    "type": "location",
    "latitude": -25.363,
    "longitude": 130.044,
    "title": "",
    "address": "Uluru (Ayers Rock)"
  }
}
```

### DIFF
--- a/packages/components/src/renderer/location.tsx
+++ b/packages/components/src/renderer/location.tsx
@@ -2,7 +2,12 @@ import { Status, Wrapper } from '@googlemaps/react-wrapper'
 import React, { useEffect, useRef } from 'react'
 import { MessageTypeHandlerProps } from '../typings'
 
-export const MapContainer: React.FC<MessageTypeHandlerProps<'location'>> = ({ latitude, longitude, address }) => {
+export const MapContainer: React.FC<MessageTypeHandlerProps<'location'>> = ({
+  latitude,
+  longitude,
+  address,
+  title
+}) => {
   const ref = useRef<HTMLDivElement>(null)
 
   useEffect(() => {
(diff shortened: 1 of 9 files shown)


packages/components/src/renderer/location.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+  return <div ref={ref} id="map" />
+}
+
+const Fallback = ({ latitude, longitude }: { latitude: number; longitude: number }): React.ReactElement => {
+  const link = `https://www.google.com/maps/search/?api=1&query=${latitude},${longitude}`
+  return (
+    <p>
+      Location:&nbsp;
+      <a href={link} target="_blank">
```
samuelmasse, line 39: I wouldn't put Location: as a fallback. It's not translatable and also it doesn't align with what other channels do. Twilio, whatsapp and others for example puts the title and adresse and the link afterwards.

packages/components/src/renderer/location.tsx
```diff
@@ -0,0 +1,82 @@
...
+      <Wrapper apiKey={config.googleMapsAPIKey} render={render(latitude, longitude)}>
+        <MapContainer
+          latitude={latitude}
+          longitude={longitude}
+          address={address}
+          title={title}
+          type={type}
+          config={config}
+        />
+      </Wrapper>
```
samuelmasse, line 77: You supply the address and title here but it doesn't show up in the webchat. Is it possible to make these visible to the webchat user?

samuelmasse (approved):
Tested and working :+1: 

Don't forget that the title and address are also parts of the location message. We show them to the user on every channel, so it would make sense to show them in the webchat also. I'm also not a fan of the hardoced `Location:` prefix for the fallback

laurentlp:
**With a valid API key**

Screenshot from 2022-03-16 11-46-09

**Fallback**

Screenshot from 2022-03-16 11-45-39

PR #411, opened 2022-03-11
chore(clients): move provider id to provisions

samuelmasse:
Preparation to use framework. Moves the providerId column from msg_clients to a new table called msg_provisions

Closes DEV-2388


samuelmasse:
CI is being annoying. Since this PR doesn't raise the server version to 1.1.5 the migration test doesn't work. We'll need to figure out a way to upgrade the server version in dev without making a release

laurentlp (approved):
LGTM

samuelmasse:
@laurentlp tested on postgres and everything works correctly

PR #414, opened 2022-03-17
chore(ci_cd): added action to create release

laurentlp:
This PR adds an action ("Create Release Pull Request") whose task is to automate the creation of the release PR. It also allows us to bump the version of the root package.json while working in the next version without automatically releasing a new binary. 

Also closes DEV-2233

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -5,7 +5,7 @@ on:
       - master
 jobs:
   release_bin:
-    if: startsWith(github.event.head_commit.message, 'release/v') == true
+    if: "${{ startsWith(github.event.head_commit.message, 'chore(server): release v') == true }}"
     name: Release Binaries
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -5,6 +5,7 @@ on:
       - master
 jobs:
   release_bin:
+    if: startsWith(github.event.head_commit.message, 'release/v') == true
     name: Release Binaries
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -5,6 +5,7 @@ on:
       - master
 jobs:
   release_bin:
+    if: "${{ startsWith(github.event.head_commit.message, 'chore(server): release v') == true }}"
```
laurentlp, line 8: It basically means that we only create a new release when the last commit message on master starts with `chore(server): release v`. Note that only the Create Release Pull Request workflow creates those commits.

.github/workflows/release.yml
```diff
@@ -16,16 +17,13 @@ jobs:
           node-version-file: '.nvmrc'
           cache: 'yarn'
 
-      - name: Git Unshallow
-        run: git fetch --prune --unshallow
```
laurentlp, line 20: Unnecessary since the `Get Release Details` already does so.

.github/workflows/release.yml
```diff
@@ -36,13 +34,11 @@ jobs:
           echo "Latest Tag: ${{ steps.release_details.outputs.latest_tag }}"
 
       - name: Build and Package
-        if: ${{ steps.release_details.outputs.is_new_release == 'true' }}
         run: |
           yarn build
           yarn package
 
       - name: Release
-        if: ${{ steps.release_details.outputs.is_new_release == 'true' }}
```
laurentlp, line 45: Those two ifs were replaced by the one at the beginning of the file.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@botpress/messaging",
-  "version": "1.1.4",
+  "version": "1.1.5",
```
laurentlp, line 3: I bumped the version here since we are working on v1.1.5

laurentlp:
@samuelmasse I tested those actions locally using `act` and most of them were inspired by actions on the main repo. Also, we will be able to remove the release script from the repo once we know this is stable.

samuelmasse (approved):

PR #418, opened 2022-03-18
chore(e2e): added basic e2e test for the webchat

laurentlp:
This PR adds the setup for running E2E tests on the webchat and adds a basic test that loads the webchat, configures it, and sends payloads.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,9 @@
     "test:int": "jest -c ./test/jest.integration.config.ts",
     "test:e2e": "jest -c ./test/jest.e2e.config.ts",
     "test:sec": "jest -c ./test/jest.security.config.ts",
-    "test:mig": "jest -c ./test/jest.migration.config.ts"
+    "test:mig": "jest -c ./test/jest.migration.config.ts",
+    "cypress:open": "cypress open",
+    "cypress:run": "cross-env HEIGHT=1080 WIDTH=1920 cypress run"
   },
   "bin": "./packages/server/dist/index.js",
   "pkg": {
@@ -62,6 +64,7 @@
     "@typescript-eslint/parser": "^5.10.2",
     "conventional-changelog": "^3.1.25",
     "cross-env": "^7.0.3",
+    "cypress": "^9.5.2",
     "eslint": "^8.8.0",
     "eslint-config-prettier": "^8.3.0",
     "eslint-import-resolver-node": "^0.3.6",
--- a/package.json
+++ b/package.json
@@ -22,8 +22,8 @@
     "test:e2e": "jest -c ./test/jest.e2e.config.ts",
     "test:sec": "jest -c ./test/jest.security.config.ts",
     "test:mig": "jest -c ./test/jest.migration.config.ts",
-    "cypress:open": "cypress open",
-    "cypress:run": "cross-env HEIGHT=1080 WIDTH=1920 cypress run"
+    "test:chat": "cross-env HEIGHT=1080 WIDTH=1920 cypress run",
+    "cypress:open": "cypress open"
   },
   "bin": "./packages/server/dist/index.js",
   "pkg": {
--- a/packages/server/src/api.ts
+++ b/packages/server/src/api.ts
@@ -64,13 +64,7 @@ export class Api {
       cors({
         origin: '*',
         methods: ['GET', 'POST', 'DELETE'],
-        allowedHeaders: [
-          'Content-Type',
-          'Authorization',
-          'Content-Length',
-          'x-bp-messaging-client-id',
-          'x-bp-messaging-client-token'
-        ]
+        allowedHeaders: '*'
       })
     )
     this.router.use(express.json({ limit: '100kb' }))
--- a/packages/server/src/api.ts
+++ b/packages/server/src/api.ts
@@ -64,7 +64,13 @@ export class Api {
       cors({
         origin: '*',
         methods: ['GET', 'POST', 'DELETE'],
-        allowedHeaders: ['Content-Type', 'Authorization', 'Content-Length']
+        allowedHeaders: [
+          'Content-Type',
+          'Authorization',
+          'Content-Length',
+          'x-bp-messaging-client-id',
+          'x-bp-messaging-client-token'
+        ]
       })
     )
     this.router.use(express.json({ limit: '100kb' }))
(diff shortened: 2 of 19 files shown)


packages/server/src/api.ts
```diff
@@ -64,7 +64,13 @@ export class Api {
       cors({
         origin: '*',
         methods: ['GET', 'POST', 'DELETE'],
-        allowedHeaders: ['Content-Type', 'Authorization', 'Content-Length']
+        allowedHeaders: [
+          'Content-Type',
+          'Authorization',
+          'Content-Length',
+          'x-bp-messaging-client-id',
+          'x-bp-messaging-client-token'
```
laurentlp, line 72: Had to add those headers there otherwise, I would have cors issues when running the tests locally.

packages/webchat/src/core/socket.tsx
```diff
@@ -10,6 +10,8 @@ export default class BpSocket {
   constructor(config: Config) {
     this.chatId = config.chatId
     this.socket = new MessagingSocket({ url: config.messagingUrl, clientId: config.clientId })
+
+    window.websocket = this.socket
```
laurentlp, line 14: Not 100% sure about this, but I haven't found any other clean way of accessing this class inside the tests.

packages/webchat/test/app.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+  useEffect(() => {
+    const config = new URLSearchParams(search).get('config')
+
+    if (!config) {
+      return
+    }
+
+    setConfig(JSON.parse(config))
+  }, [search])
```
laurentlp, line 21: Watching changes on the query params allows us to update the webchat config during the test so that we can easily test different configs.

packages/server/src/api.ts
samuelmasse: Can we just allow any header? Why would we need this setting anyway?

samuelmasse: Can we just allow any header? Why would we need this setting anyway?

test/cypress/plugins/index.ts
```diff
@@ -0,0 +1,22 @@
+/// <reference types="cypress" />
+// ***********************************************************
+// This example plugins/index.js can be used to load plugins
+//
+// You can change the location of this file or turn off loading
+// the plugins file with the 'pluginsFile' configuration option.
+//
```
samuelmasse, line 7: Are the files in the test/cypress folder necessary? They don't seem to be doing anything

package.json
```diff
@@ -21,7 +21,9 @@
     "test:int": "jest -c ./test/jest.integration.config.ts",
     "test:e2e": "jest -c ./test/jest.e2e.config.ts",
     "test:sec": "jest -c ./test/jest.security.config.ts",
-    "test:mig": "jest -c ./test/jest.migration.config.ts"
+    "test:mig": "jest -c ./test/jest.migration.config.ts",
+    "cypress:open": "cypress open",
+    "cypress:run": "cross-env HEIGHT=1080 WIDTH=1920 cypress run"
```
samuelmasse, line 26: What are those for? Can we have a yarn test:webchat

packages/webchat/src/declaration.d.ts
```diff
@@ -0,0 +1 @@
+declare module '*.scss'
```
samuelmasse, line 1: There's already this line in globals.d.ts. Why do we need this file also?

packages/webchat/test/tsconfig.json
```diff
@@ -0,0 +1,18 @@
...
+
+    "jsx": "react",
+    "lib": ["dom", "es2021"],
+
+    // be explicit about types included
+    // to avoid clashing with Jest types
+    "types": ["cypress"]
+  },
+  "include": ["../../../node_modules/cypress", "../test/**/*", "../src"]
+}
```
samuelmasse, line 18: The include here normally lists `"../src/**/*"` not `"../src"`. Is there any reason for it being different?

packages/webchat/test/tsconfig.json
```diff
@@ -0,0 +1,18 @@
+{
+  "extends": "../../../tsconfig.packages.json",
+  "compilerOptions": {
+    "noEmit": true,
```
samuelmasse, line 4: Is noEmit needed? We don't have it in the other test tsconfig

samuelmasse (approved):
Tested and working :+1: . Way better than jest good job! I left a few comments other than it everything looks good

packages/server/src/api.ts
laurentlp: I think it's good practice to have those there. See that for reference: https://stackoverflow.com/questions/17992042/whats-the-point-of-the-access-control-allow-headers-header

But in our case, since no one will really make requests to messaging from the front-end, we may consider removing them.

package.json
laurentlp: This was recommended by the cypress doc, but yes we can. What do you suggest?

samuelmasse: I suggest we have a one liner way of starting the server, starting the webchat and running the tests. A script could do the job or anything else. The process of running the e2e in botpress is way too complicated and it would be better to have it streamlined here

samuelmasse: yarn test:chat

It would be awesome if it could display the tests it's doing automatically aswell

packages/webchat/src/declaration.d.ts
laurentlp: I moved this line into this file because it was clashing with the other module I declare in the global typings.

packages/webchat/test/tsconfig.json
laurentlp: We don't need to include anything from the source except the typings (`"../src"`), the one from cypress is recommended by them for some reason, and lastly, we include all the test files.

packages/webchat/test/tsconfig.json
laurentlp: Not sure if it is needed, but, again this is a recommended step by the Cypress documentation

test/cypress/plugins/index.ts
laurentlp: This file was autogenerated, so I did not touch it.

packages/server/src/api.ts
samuelmasse: Yup I'd just remove them if they're not required

PR #419, opened 2022-03-21
chore(tests): add migration diff tests

laurentlp:
This PR adds the basis for the migration diff test which performs up and down operations on migrations to make sure of the schema integrity stays intact when running migrations.


packages/server/test/migration/migrations-diff.test.ts
```diff
@@ -0,0 +1,70 @@
...
+
+describe('Migrations diff tests', () => {
+  const migrationFolder = path.join(__dirname, '../../src/migrations')
+  const migrationFiles = fs.readdirSync(migrationFolder)
+  const migrations = migrationFiles
+    .filter((f) => !isNaN(Number(f.charAt(0)))) // Only keep migration files
+    .map((f) => {
+      const version = f.split('-')[0]
+      return [f, version, decrement(version)]
+    })
```
samuelmasse: Don't need to do this. The migrations are listed in the packages/server/serc/migrations/index.ts file. You can just import the array from there. The file names of migrations are not significative (by design). They only have the name of the migration and the version for organizational purposes. There is a `meta` property of the base migration class that you can use to get this info.

samuelmasse: To get instances you can do the same thing the MigrationService does with `const all = this.migs.map((x) => new x())`

package.json
```diff
@@ -21,7 +21,7 @@
     "test:int": "jest -c ./test/jest.integration.config.ts",
     "test:e2e": "jest -c ./test/jest.e2e.config.ts",
     "test:sec": "jest -c ./test/jest.security.config.ts",
-    "test:mig": "jest -c ./test/jest.migration.config.ts"
+    "test:mig": "yarn build && jest -c ./test/jest.migration.config.ts"
   },
```
samuelmasse, line 25: Why build? You can use yarn dev for the migration commands also

packages/server/test/migration/migrations-diff.test.ts
```diff
@@ -0,0 +1,70 @@
...
+    const afterMigrationDatabase = `after_mig_${migrationName}`
+    const beforeMigrationDatabase = `before_mig_${migrationName}`
+
+    test(
+      'Should run each migrations up to the migration we are testing',
+      async () => {
+        await startMessagingServer(
+          {
+            command: `yarn start migrate up --target ${previousVersion}`,
+            launchTimeout: TIMEOUT
```
samuelmasse, line 36: Why use `yarn start`? Was something not working with `yarn dev`?

samuelmasse (commented):
Looks like it's mostly working (i think the issue with the CI is that both tests try to use the 3001 port).

The list of migrations to check is available as an array in the packages/server/src/migrations/index.ts file. So no need to do complicated file listing. I think this file listing might be the reason you had to use `yarn build` so getting rid of that should allow use to run the test without building everything.

Something I noticed is that all the migration tests are run sequentially. This won't scale well in the future so we should try to get them to run in parallel like other tests. Usually tests in different files run in parallel, so I don't know how that would work with the current setup of generating the tests in one file.

package.json
laurentlp: Since we call yarn dev like 30 times, it simply runs faster if we build the project first.

packages/server/test/migration/migrations-diff.test.ts
laurentlp: It's just faster

packages/server/test/migration/migrations-diff.test.ts
laurentlp: Right, I'll adapt this.

package.json
samuelmasse: Makes sense from a CI point of view, but if you're in the process of writing a migration it's annoying to have to build the project to run the test

samuelmasse: Sometimes I will do a yarn test to check if everything is still working correctly. I test:mig builds the project it's going to be way longer

laurentlp: It's faster overall, that's the reason why I made this change. I'll revert it since I don't really care.

samuelmasse (approved):
LGTM

PR #423, opened 2022-03-22
chore(tests): add migration cli tests

laurentlp:
Self-explanatory. We cover all the unhappy and happy paths that a user can take when interfacing with the migration CLI.

Here are the options: 

- `--auto-migrate`
- `migrate <up|down> [--target <target>] [--dry]`

### DIFF
--- a/packages/server/test/migration/global-diff.test.ts
+++ b/packages/server/test/migration/global-diff.test.ts
@@ -1,3 +1,5 @@
+import portfinder from 'portfinder'
+
 import { compareDatabases } from './utils/diff'
 import { startMessagingServer } from './utils/server'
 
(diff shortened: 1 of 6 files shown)


packages/server/test/migration/global-diff.test.ts
```diff
@@ -9,13 +11,14 @@ describe('Global Diff', () => {
   test(
     'Starts Messaging with the latest database schema',
     async () => {
+      const port = await portfinder.getPortPromise()
       await startMessagingServer(
         {
```
samuelmasse, line 16: I think we should use portfinder on all of our tests. I regularly run the tests when some instance of messaging is running at it forces be to close that instance of messaging

PR #425, opened 2022-03-24
feat(channels): custom channels

samuelmasse:
Adds a way to create mappings for any arbitrary string. The channel part of the endpoint doesn't have to be a predefined channel anymore

For a full demo of how to use this feature see : https://github.com/botpress/custom-channel

Closes DEV-2397

### DIFF
--- a/packages/base/src/endpoint.ts
+++ b/packages/base/src/endpoint.ts
@@ -1,5 +1,5 @@
 export interface Endpoint {
-  channel: { name: string; version: string }
+  channel: string | { name: string; version: string }
   identity: string
   sender: string
   thread: string
--- a/packages/client/src/schema.ts
+++ b/packages/client/src/schema.ts
@@ -33,7 +33,7 @@ export const Schemas = {
         .object({
           id: joi.string().uuid().required(),
           conversationId: joi.string().uuid().required(),
-          authorId: joi.string().uuid().required(),
+          authorId: joi.string().uuid().optional(),
           sentOn: joi.date().required(),
           payload: joi.object().required()
         })
--- a/packages/server/src/base/streamer.ts
+++ b/packages/server/src/base/streamer.ts
@@ -67,15 +67,14 @@ export class Streamer {
       await this.dispatcher.publish(StreamerDispatches.Message, userId, { source: source?.socket?.id, payload })
     }
 
-    if (source?.client?.id !== clientId) {
-      const webhooks = await this.webhooks.list(clientId)
+    // TODO: this might break stuff on botpress's side
+    const webhooks = await this.webhooks.list(clientId)
 
-      for (const webhook of webhooks) {
-        void this.send(process.env.SPINNED_URL || webhook.url, payload, {
-          'x-bp-messaging-client-id': clientId,
-          'x-bp-messaging-webhook-token': webhook.token
-        })
-      }
+    for (const webhook of webhooks) {
+      void this.send(process.env.SPINNED_URL || webhook.url, payload, {
+        'x-bp-messaging-client-id': clientId,
+        'x-bp-messaging-webhook-token': webhook.token
+      })
     }
   }
 
--- a/packages/server/src/base/streamer.ts
+++ b/packages/server/src/base/streamer.ts
@@ -67,7 +67,6 @@ export class Streamer {
       await this.dispatcher.publish(StreamerDispatches.Message, userId, { source: source?.socket?.id, payload })
     }
 
-    // TODO: this might break stuff on botpress's side
     const webhooks = await this.webhooks.list(clientId)
 
     for (const webhook of webhooks) {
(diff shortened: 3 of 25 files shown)


packages/server/src/base/streamer.ts
```diff
@@ -67,15 +67,14 @@ export class Streamer {
...
-      }
+    // TODO: this might break stuff on botpress's side
+    const webhooks = await this.webhooks.list(clientId)
+
+    for (const webhook of webhooks) {
+      void this.send(process.env.SPINNED_URL || webhook.url, payload, {
+        'x-bp-messaging-client-id': clientId,
+        'x-bp-messaging-webhook-token': webhook.token
+      })
     }
```
samuelmasse, line 78: I think we'll have to ignore bot message on botpress

packages/base/src/endpoint.ts
```diff
@@ -1,5 +1,5 @@
 export interface Endpoint {
-  channel: { name: string; version: string }
+  channel: string | { name: string; version: string }
   identity: string
```
samuelmasse, line 3: If the channel value is a string then it's a custom channel. If it's an object with a version and name then it's a builtin channel

packages/client/src/schema.ts
```diff
@@ -33,7 +33,7 @@ export const Schemas = {
         .object({
           id: joi.string().uuid().required(),
           conversationId: joi.string().uuid().required(),
-          authorId: joi.string().uuid().required(),
+          authorId: joi.string().uuid().optional(),
           sentOn: joi.date().required(),
```
samuelmasse, line 37: The schema for messages was incorrect. A message can have a null author (bot message)

packages/base/src/endpoint.ts
laurentlp: Why not just add a custom property or something?

packages/server/src/base/streamer.ts
laurentlp: And why is that?

packages/server/src/mapping/tunnels/service.ts
```diff
@@ -90,6 +117,23 @@ export class TunnelService extends Service {
     }
   }
 
+  private async getByClientAndCustomChannel(clientId: uuid, customChannelName: string): Promise<Tunnel | undefined> {
+    const cached = this.cacheByClientAndCustomChannel.get(clientId, customChannelName)
+    if (cached) {
+      return cached
+    }
+
+    const rows = await this.query().where({ clientId, customChannelName })
```
laurentlp, line 126:
```suggestion
    const rows = await this.query().where({ clientId, customChannelName }).first()
```

packages/server/src/migrations/1.1.7-custom-channels.ts
```diff
@@ -0,0 +1,33 @@
+import { getTableId, Migration } from '@botpress/messaging-engine'
+
+export class CustomChannelsMigration extends Migration {
+  meta = {
+    name: CustomChannelsMigration.name,
+    description: 'Modifies the msg_tunnels table to support custom channels',
+    version: '1.1.5'
```
laurentlp, line 7:
```suggestion
    version: '1.1.7'
```

packages/server/src/base/streamer.ts
samuelmasse: We don't want the bot reacting to messages it just posted

packages/base/src/endpoint.ts
samuelmasse: It's for simplicity. You're rarely ever going to do custom mapping for a builtin channel (can't even think of a scenario where that would make sense), so for 99.9% of use cases it's simpler for it to just be a string

packages/server/src/mapping/tunnels/service.ts
samuelmasse: Not putting first here is consistent with the rest of the codebase. Anyways this query isn't supposed to return more than one row

packages/server/src/migrations/1.1.7-custom-channels.ts
samuelmasse: Good catch

packages/server/src/base/streamer.ts
laurentlp: Then why not extend the mechanism that tells the source of a message so that we know not to send the event?

samuelmasse: We'll have that with more complex webhook settings (only send incoming/outgoing, only send for x channel, etc). To be done in future PR

laurentlp (approved):

fgrep:
Hello @samuelmasse. Can you provide the full demo for this feature? The link https://github.com/botpress/custom-channel does not exist. Thanks

samuelmasse:
Hey @fgrep I changed the repo visibility to public, you can refer to it to implement your own custom channel. You can use bp.messaging that was added to the skd in 12.27

PR #428, opened 2022-03-28
feat(user-tokens): user tokens api

samuelmasse:
Adds API routes to create user tokens

Closes DEV-2308

### DIFF
--- a/packages/client/src/api.ts
+++ b/packages/client/src/api.ts
@@ -86,6 +86,16 @@ export abstract class MessagingChannelApi extends MessagingChannelBase {
   }
 
   /**
+   * Creates a new user token
+   * @param clientId id of the client that owns the user
+   * @param userId id of the user
+   * @returns token that can be used to make user-level requests
+   */
+  async createUserToken(clientId: uuid, userId: uuid): Promise<string> {
+    return (await this.http.post('/users/tokens', { userId }, { headers: this.headers[clientId] })).data.token
+  }
+
+  /**
    * Creates a new messaging conversation
    * @param clientId id of the client that will own this conversation
    * @param userId id of the user that starts this conversation
--- a/packages/server/src/user-tokens/api.ts
+++ b/packages/server/src/user-tokens/api.ts
@@ -1,0 +1,29 @@
+import { uuid } from '@botpress/messaging-base'
+import { Response } from 'express'
+import { ApiManager } from '../base/api-manager'
+import { ClientApiRequest } from '../base/auth/client'
+import { UserService } from '../users/service'
+import { Schema } from './schema'
+import { UserTokenService } from './service'
+
+export class UserTokenApi {
+  constructor(private users: UserService, private userTokens: UserTokenService) {}
+
+  setup(router: ApiManager) {
+    router.post('/users/tokens', Schema.Api.Create, this.create.bind(this))
+  }
+
+  async create(req: ClientApiRequest, res: Response) {
+    const userId = req.body.userId as uuid
+
+    const user = await this.users.fetch(userId)
+    if (!user || user.clientId !== req.clientId) {
+      return res.sendStatus(404)
+    }
+
+    const tokenRaw = await this.userTokens.generateToken()
+    const userToken = await this.userTokens.create(user.id, tokenRaw, undefined)
+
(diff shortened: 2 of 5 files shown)


packages/server/src/sync/schema.ts
```diff
@@ -14,10 +15,11 @@ export const makeSyncRequestSchema = (channels: Channel[]) => {
...
   for (const [name, channels] of Object.entries(channelsByName)) {
+    const versionSchema = Joi.string().valid(...new Set(channels.map((x) => x.meta.version)))
+
     channelsSchema[name] = Joi.object({
-      version: Joi.string()
-        .valid(...channels.map((x) => x.meta.version))
-        .optional()
+      // when legacy channels are enable, not supplying the version defaults to 0.1.0
+      version: yn(process.env.ENABLE_LEGACY_CHANNELS) ? versionSchema.optional() : versionSchema.required()
     })
```
samuelmasse, line 23: Unrelated fix to the 500 error you would get if you tried to setup a legacy channel with legacy channel disabled

packages/client/src/api.ts
```diff
@@ -85,6 +85,16 @@ export abstract class MessagingChannelApi extends MessagingChannelBase {
...
+  /**
+   * Creates a new user token
+   * @param clientId id of the client that owns the user
+   * @param userId id of the user
+   * @returns token that can be used to make user-level requests
+   */
+  async createUserToken(clientId: uuid, userId: uuid): Promise<string> {
+    return (await this.http.post('/users/tokens', { userId }, { headers: this.headers[clientId] })).data.token
+  }
+
```
samuelmasse, line 97: Note : need to publish a new version of the client with these changes

packages/server/src/user-tokens/api.ts
```diff
@@ -0,0 +1,29 @@
...
+    const user = await this.users.fetch(userId)
+    if (!user || user.clientId !== req.clientId) {
+      return res.sendStatus(404)
+    }
+
+    const tokenRaw = await this.userTokens.generateToken()
+    const userToken = await this.userTokens.create(user.id, tokenRaw, undefined)
+
+    res.status(201).send({ token: `${userToken.id}.${tokenRaw}` })
+  }
```
samuelmasse, line 28: Should we also include the id seperatly to make the delete call easier? Otherwise need to split the token string

packages/webchat/src/main.tsx
```diff
@@ -174,7 +174,16 @@ class Web extends React.Component<MainProps> {
       case 'configure':
         return this.props.updateConfig!(Object.assign({}, constants.DEFAULT_CONFIG, data.payload))
       case 'mergeConfig':
-        return this.props.mergeConfig!(data.payload)
+        this.props.mergeConfig!(data.payload)
+
+        const oldUserId = this.socket.socket.userId
+        await this.socket.reload(data.payload)
```
laurentlp, line 180: We only want to reconnect when **some** of the config changed, not for all of them

packages/server/src/user-tokens/api.ts
laurentlp: Yes let's go this way, I think it will be easier and more intuitive for the user.

packages/webchat/src/main.tsx
samuelmasse: I made it so the reload function of the socket checks if reconnection is needed

packages/client/test/e2e/client.test.ts
```diff
@@ -210,6 +210,16 @@ describe('Http Client', () => {
...
 
+  describe('User Tokens', () => {
+    test('Should create a user token without throwing any error', async () => {
+      const userToken = await client.createUserToken(state.user!.id)
+
+      expect(userToken.id).toBeDefined()
+      expect(userToken.token).toBeDefined()
+      expect(userToken.token.startsWith(userToken.id)).toBeTruthy()
+    })
+  })
```
laurentlp, line 221: You could maybe add another test that makes sure we can create more than one user token for a single user?

laurentlp (approved):
Small comment otherwise LGTM

packages/client/test/e2e/client.test.ts
samuelmasse: Sure

PR #432, opened 2022-04-01
chore(messaging): use framework even more

samuelmasse:
Goes with  #431. Uses the launcher of the framework for the messaging server

Todo (future PR): 
- Bring back list of channels when server starts
- framework should export everything from engine, and server should import stuff only from framework

Closes DEV-2548


packages/framework/src/launcher.ts
```diff
@@ -68,15 +68,20 @@ export class Launcher {
...
 
-      this.logger.info(`Server is listening at: http://localhost:${port}`)
-
-      const externalUrl = process.env.EXTERNAL_URL
-      if (externalUrl?.length) {
-        this.logger.info(`Server is exposed at: ${externalUrl}`)
+      // TODO: remove mentions of messaging here
+      if (!yn(process.env.SPINNED)) {
+        this.logger.info(`Server is listening at: http://localhost:${port}`)
+
```
samuelmasse, line 74: In future PR

packages/server/src/sync/service.ts
```diff
@@ -49,6 +52,18 @@ export class SyncService extends Service {
...
+    const provision = await this.provisions.fetchByClientId(clientId)
+    if (provision) {
+      return provision.providerId
+    }
+
+    const provider = await this.providers.create(clientId, false)
+    await this.provisions.create(clientId, provider.id)
+
+    return provider.id
+  }
```
samuelmasse, line 65: We need to do this now because the provider isn't created when the client is created anymore

packages/server/test/integration/client-tokens.test.ts
```diff
@@ -24,7 +23,7 @@ describe('ClientTokens', () => {
   })
 
   afterAll(async () => {
-    await app.destroy()
+    await destroyApp()
```
samuelmasse, line 26: Destroying the app now requires 3 steps so I put it in its own function

laurentlp (approved):

PR #436, opened 2022-04-05
chore(tests): use services instead of cli

laurentlp:
This PR improves the migration tests by using services directly instead of using the CLI for each test. This improves the speed of global and migration diff tests by a factor of at least 15. 

I also improve the time it takes to run the CLI test by having a step that builds the server before running. This change removes the unnecessary call to `yarn start` each time we interact with the CLI.


samuelmasse (approved):
Tested and working :+1:

PR #438, opened 2022-04-06
chore(tests): improve database test setup

laurentlp:
This PR improves the database test setup by using the database server and creating all the required databases instead of having many tables suffixed with a certain id in the same database.

This PR will allow me to easily clone PostgreSQL databases in some of our test suites.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -52,7 +52,7 @@ jobs:
         run: |
           yarn test:int --coverage --verbose
         env:
-          DATABASE_URL: ${{ matrix.database == 'postgres' && 'postgres://postgres:postgres@localhost:5432/postgres' || '' }}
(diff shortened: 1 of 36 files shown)


.github/workflows/tests.yml
```diff
@@ -52,7 +52,7 @@ jobs:
         run: |
           yarn test:int --coverage --verbose
         env:
-          DATABASE_URL: ${{ matrix.database == 'postgres' && 'postgres://postgres:postgres@localhost:5432/postgres' || '' }}
```
laurentlp, line 55: This change allows us to create databases and simply have to append its name to the connection string to communicate with it and it also fallback to the default database (`postgres`) if used directly.

samuelmasse (approved):
LGTM. We need to replace the stuff like ${tableName}.column and also the databases need to be deleted after the tests are run

PR #440, opened 2022-04-07
chore(tests): add migration tests with data

laurentlp:
This PR brings the first few steps of having seed data on our migration tests.


package.json
```diff
@@ -16,12 +16,19 @@
     "dev": "yarn workspace @botpress/messaging-server dev",
     "board": "yarn workspace @botpress/messaging-board dev",
     "channels": "yarn workspace @botpress/messaging-channels example",
+    "test:setup": "docker run -p 2345:5432 --rm --name msg_test_postgres -d -e POSTGRES_USER=postgres -e POSTGRES_PASSWORD=postgres -e POSTGRES_DB=postgres postgres && sleep 2",
+    "test:teardown": "docker container stop msg_test_postgres",
     "test": "yarn test:unit && yarn test:int && yarn test:mig && yarn test:e2e && yarn test:sec",
+    "test:pg": "yarn test:unit && yarn test:int:pg && yarn test:mig:pg && yarn test:e2e:pg && yarn test:sec:pg",
```
samuelmasse, line 22: Any reason we can't have test:setup and test:teardown inside this command?

package.json
```diff
@@ -16,12 +16,19 @@
     "dev": "yarn workspace @botpress/messaging-server dev",
     "board": "yarn workspace @botpress/messaging-board dev",
     "channels": "yarn workspace @botpress/messaging-channels example",
+    "test:setup": "docker run -p 2345:5432 --rm --name msg_test_postgres -d -e POSTGRES_USER=postgres -e POSTGRES_PASSWORD=postgres -e POSTGRES_DB=postgres postgres && sleep 2",
+    "test:teardown": "docker container stop msg_test_postgres",
```
samuelmasse, line 20: As this is strictly postgres, why not call it test:setup:pg ?

samuelmasse (approved):
LGTM

I think it would be a good idea to seperate the seeded and non seeded tests. The migration "diff" test is really a test about schemas, so I would have another test that does the exact same thing but with seeded data and would name it like "up down" test

There is still a problem with postgres where running a command twice will have the second run run on databases that were corrupted by the previous test. You would need to do setup and teardown between each run to avoid this problem (for example you can try running test:mig:pg twice and the second run fails)

packages/server/test/migration/migrations-diff.test.ts
```diff
@@ -51,9 +73,13 @@ describe('Migrations diff tests', () => {
       'Should run the migration without error',
       async () => {
         process.env.MIGRATE_CMD = 'up'
+        process.env.TESTMIG_DB_VERSION = previousVersion
```
samuelmasse, line 76: What is that for?

package.json
laurentlp: I added it at first, but then considering that we don't do it for each sub-command, I thought it would be better to just have to do it all the time. Would you prefer it if those commands were added here?

PR #444, opened 2022-04-12
fix(shutdown): fix channels not being cleared correctly

samuelmasse:
Fixes an issue where conduits that are currently running wouldn't be cleared up when the server closes

Closes DEV-2566

### DIFF
--- a/packages/server/src/instances/clearing/service.ts
+++ b/packages/server/src/instances/clearing/service.ts
@@ -34,6 +34,11 @@ export class InstanceClearingService extends Service {
 
     for (const channel of this.channels.list()) {
       channel.stateManager({
+        scopes: () =>
+          this.statesCache
+            .keys()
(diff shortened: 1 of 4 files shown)


packages/server/src/instances/clearing/service.ts
```diff
@@ -34,6 +34,11 @@ export class InstanceClearingService extends Service {
 
     for (const channel of this.channels.list()) {
       channel.stateManager({
+        scopes: () =>
+          this.statesCache
+            .keys()
+            .filter((x) => x.startsWith(channel.meta.id))
+            .map((x) => this.statesCache.getValues(x)[1]),
```
laurentlp, line 41: Maybe we should consider adding something better than `any` as the typing for the `statesCache`. That way, it would also make more sense what you access here!

laurentlp (approved):
Small comment else LGTM!

PR #445, opened 2022-04-19
chore(socket): don't log every socket event

samuelmasse:
Prevents everything happening with sockets from being logged to the console

Also adds ENABLE_BILLING_STATS to get logging for amount of messages received, sent even if BILLING_ENDPOINT is not set

Closes DEV-2583


laurentlp (approved):
LGTM!

laurentlp (approved):
LGTM!

PR #447, opened 2022-04-22
chore(tests): improved running postgresql tests locally

laurentlp:
This PR greatly improves the setup for running tests using PostgreSQL. 

Now, instead of having to manually call `yarn test:<setup|teardown>:pg` before running tests, you simply have to run any testing script as you would normally do, but suffix it with `:pg` to run with postgres. The container will be created automatically and removed at the end of each test.

If you prefer running your own PostgreSQL database, simply call the "normal" test script (e.g. `yarn test:unit`) while setting the `DATABASE_URL` env var to point to your local instance. That's what we are doing on the CI.


samuelmasse (approved):
This improves the setup by removing the setup and teardown step. However it doesn't fix the issue that you can't run `yarn test:pg` twice. The data created from the previous run corrupts the second run

laurentlp:
> This improves the setup by removing the setup and teardown step. However it doesn't fix the issue that you can't run `yarn test:pg` twice. The data created from the previous run corrupts the second run

This is not true! You can actually run the same command many times in a row without issue.

PR #450, opened 2022-04-25
chore(yarn): use pnp

samuelmasse:
Converts the repo to use pnp. Removes legacy channels and pkg. Branche `v1.1` was created to do updates to the old version of messaging

Closes DEV-2603


packages/webchat/test/index.html
```diff
@@ -5,7 +5,10 @@
     <title>Webchat</title>
 
     <link rel="stylesheet" href="../../components/src/css/botpress-default.css" />
-    <link rel="stylesheet" href="../../../node_modules/@blueprintjs/core/lib/css/blueprint.css" />
+    <link
+      rel="stylesheet"
+      href="../../.././.yarn/cache/@blueprintjs-core-npm-3.52.0-8e614f339f-e0d46e0446.zip/node_modules/@blueprintjs/core/lib/css/blueprint.css"
```
laurentlp, line 10: I think you should unplug blueprint instead of doing this: https://yarnpkg.com/cli/unplug

laurentlp (approved):
Small comment otherwise LGTM!

packages/webchat/test/index.html
samuelmasse: not yet

PR #452, opened 2022-04-25
chore(deps): bumps parcel and storybook

laurentlp:
This PR bumps Parcel and Storybook to their latest versions.


samuelmasse:
Bumping parcel does not work at the moment due to several problems. We'll see what we can do about it after updating the repo to use PnP

PR #453, opened 2022-04-25
chore: restructure repo

samuelmasse:
Reorganizes repo with "super directories" that contains groups of packages seperated by concern (we'll have a `base` group, a `messaging` group and in the future a `studio` group and a `runtime` group) 

Closes DEV-2604


samuelmasse:
Note for future PR :
- Move webchat related packages into their own group
- Move cypress.json to the test folder
- Move the dockerfile and dockerignore for the messaging server to the location of the package

PR #456, opened 2022-04-26
chore: refact inspector kit test and storybook setup

samuelmasse:
We'll make it work the same way as the components package

Note : don't know how the tests should be run

Closes DEV-2616


packages/studio/inspector-kit/story/tsconfig.json
```diff
@@ -0,0 +1,17 @@
...
+    // We use react
+    "jsx": "react",
+    "resolveJsonModule": true
+  },
+  "include": [
+    "../story/**/*",
+    "../test/**/*.json",
+    "../src/**/*",
+    // this should probably not be there
+    "../demo/**/*"
```
laurentlp, line 15: Then remove it, same thing for the test files and we'll re-add them if necessary

.gitignore
```diff
@@ -18,3 +18,6 @@ test/cypress/videos
 test/cypress/fixtures
 test/cypress/screenshots
 test/cypress/downloads
+.rollup.cache
+lib
+node_modules
```
laurentlp, line 23: Do we still have node_modules?

packages/studio/inspector-kit/story/tsconfig.json
samuelmasse: It needs to be there because we import stuff from the demo, this is just a note that this should be removed later

laurentlp (approved):

PR #461, opened 2022-04-28
chore(ci_cd): improve ci to accommodate all packages

laurentlp:
Description coming soon...


.github/workflows/build.yml
```diff
@@ -1,9 +1,15 @@
 name: Build
 on: [pull_request]
+
 jobs:
+  files_changed:
+    uses: 'botpress/messaging/.github/workflows/files-changed.yml@master'
```
laurentlp, line 6: Using `@master` here will make the checks fail, but will pass once this PR is merged since it adds this workflow

.github/workflows/build.yml
```diff
@@ -68,4 +75,46 @@ jobs:
...
+      - uses: actions/setup-node@v2
+        with:
+          node-version-file: '.nvmrc'
+          cache: 'yarn'
+      - name: Fetch Node Packages
+        run: |
+          yarn --immutable
+      - name: Build Messaging
+        run: |
+          yarn workspace @botpress/${{ matrix.path }} build:tsc
```
laurentlp, line 100: I think we should have the reflex of type checking all packages that use parcel.

PR #468, opened 2022-05-02
chore(studio): re-implemented export to archive util function

laurentlp:
This PR simply re-implements the export to archive buffer ghost utility function to the `Instance` util.

To test this feature, you can upload your bot to the cloud and see if it works as expected.


Raid55 (approved):
LGTM, only a conflict with yarn.lock then good to go

packages/studio/server/src/studio/nlu/nlu-service.ts
```diff
@@ -148,29 +149,26 @@ export class NLUService {
...
     const modelsFolder = 'models'
-    await this.ghost.forBot(botId).deleteFolder(modelsFolder)
+    await Instance.deleteDir(modelsFolder)
 
     for (const lang of Object.keys(botConfig.nluModels)) {
       const modelId = botConfig.nluModels[lang]
       const modelWeights = await bot.downloadModelWeights(botId, modelId)
 
-      await this.ghost.forBot(botId).upsertFile(modelsFolder, `${modelId}.model`, modelWeights)
+      await Instance.upsertFile(path.join(modelsFolder, `${modelId}.model`), modelWeights)
```
EFF, line 170: **question**: have you verified modelWeights encoding ? i seems you save it as raw bytes (or downloadModelWeights already pareses it). I imagine you tried to upload it on cloud and works.

packages/studio/server/src/studio/utils/bpfs.ts
```diff
@@ -93,5 +96,38 @@ export const Instance: bpfs = {
   },
   moveFile(fromPath: string, toPath: string): Promise<void> {
     return fse.move(resolvePath(fromPath), resolvePath(toPath))
+  },
+  async exportToArchiveBuffer(): Promise<Buffer> {
```
EFF, line 100: **nitpick**: although writing on disk is totally fine on the studio app we could have done this in all in mem. (not to be fixed)

EFF (approved):
LGTM

packages/studio/server/src/studio/nlu/nlu-service.ts
laurentlp: I haven't checked the encoding, but I have tried to upload it and it worked

PR #470, opened 2022-05-02
chore(studio): adapt media service to new conversation studio

laurentlp:
This PR:

 - Removes legacy code
 - Updates the CMS to provide a BOT_URL
 - Updates the content-type utils regex used to determine if the URL is coming from botpress
 - Removed the unnecessary usage of a reducer in the Upload component
 - Improved typings here and there
 - Removed now unused translations


Raid55 (approved):
LGTM, I am not super familiar with some of the parts touched on but it looks fine

packages/studio/server/src/studio/cms/cms-router.ts
```diff
@@ -214,8 +214,8 @@ export class CMSRouter extends CustomStudioRouter {
       }
     }
 
-    // TODO: add BOT_URL
-    const context = { BOT_ID: '' }
+    const host = process.EXTERNAL_URL || `http://localhost:${process.PORT}`
```
EFF, line 217: should we use `process.host` as well ?

EFF (approved):

packages/studio/server/src/studio/cms/cms-router.ts
laurentlp: Good point, I think we should!

PR #473, opened 2022-05-03
chore(studio): re-add testing router

laurentlp:
This PR re-adds the testing module backend to the Studio Server. 

As you will see, I added a lot of TODOs and commented out some pieces of code. This is because we will need to have some ways of querying the runtime to fetch this missing info. I don't think, for example, that the studio should have a direct connection to the runtime's database.

### DIFF
--- a/packages/studio/server/src/studio/testing/recorder.ts
+++ b/packages/studio/server/src/studio/testing/recorder.ts
@@ -1,0 +1,83 @@
+import * as sdk from '@botpress/sdk'
+import _ from 'lodash'
+
+import { Scenario } from './typings'
+import { convertLastMessages } from './utils'
+
+export class Recorder {
+  private _lastEvent?: sdk.IO.IncomingEvent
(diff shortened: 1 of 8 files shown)


packages/studio/server/src/studio/testing/recorder.ts
```diff
@@ -0,0 +1,83 @@
...
+    }
+  }
+
+  async processCompleted(event: sdk.IO.IncomingEvent): Promise<void> {
+    if (!this.isRecording() || !this._target) {
+      return
+    }
+
+    // TODO: This should most likely be a call to the runtime
+    const target = ''
```
EFF, line 33: or messaging ?

laurentlp: Right! In any case, I think we should take some time to properly rethink the testing module considering that we now have the messaging server and the runtime as standalone. Maybe we can refactor and remove a lot of logic there.

PR #477, opened 2022-05-05
chore(deps): bump dependencies

laurentlp:
This PR bumps all dependencies to their latest version that does not break our current setup.

*Notes:*
 - Jest cannot be updated to v28 since it doesn't seem to be working with ts-node 
 - Parcel cannot be updated to 2.3.X+ since there are known issues with NodeJS-related libs and blueprintjs CSS files.
 - React Redux and Slate React have not been updated since they require specific versions to work well with other deps.
 - All other dependencies have been updated to their closest minor version


Raid55 (approved):
looks good

PR #479, opened 2022-05-07
chore(tests): major refactor and improvements around our test setup

laurentlp:
Sorry guy, this is a big one...

What I did was create a base test package that contains reusable utils that we copy-pasted across all our test suites. Also, all the setup and teardown files were shared and defined at the root of the repo. Since we now have many servers in the same repo (messaging, studio, and runtime), we don't want to be sharing the same setup for each package. That's why I moved those setup files under each package's test folder (e.g. `packages/messaging/server/test/jest.e2e.setup.ts` instead of being at `test/jest.e2e.setup.ts`). I also increased the timeouts for when we wait for the servers to start so that when the CI is slow, it doesn't fail while waiting for messaging to start.


packages/base/test/src/database/sqlite.ts
```diff
@@ -0,0 +1,18 @@
...
+export const getTestDataPath = () => {
+  return path.resolve(__dirname, '../../../../../test/.test-data')
+}
+
+export const setup = () => {
+  process.env.DATABASE_URL = process.env.DATABASE_URL || path.join(getTestDataPath(), `${uuidv4()}.sqlite`)
+}
+
+export const teardown = () => {
+  const dir = path.join(getTestDataPath())
```
EFF, line 14: please leave a comment on why we don't tear down `process.env.DATABASE_URL`

EFF (approved):
good work

PR #481, opened 2022-05-11
chore(inspector): clean up scss for kit compoennts

sidcarrollworks:


laurentlp:
> Not sure what's happening in the CI here. The build studio is failing, but it's outputting errors for the frontend code not the backend @laurentlp ?

I have no clue why no. What we run on the CI is: `yarn workspace @botpress/studio-be build:tsc` and `build:tsc` is a shortcut for `yarn run -T tsc --build`. Furthermore, if I then look in the tsconfig, we only include the src folder and exclude the dist.

PR #484, opened 2022-05-12
chore(studio): add boilerplate for conversation studio

laurentlp:
This PR contains the boilerplate for the new Conversation Studio that contains examples on how to use libraries like:

 - React v17 (we cannot use v18 yet since we have resolutions in the main package.json blocking us)
 - Zustand v4
 - React-Router v4
 - React-Query v3
 - React-i18next v11
 - Parcel v2


packages/studio/frontend2/src/App.tsx
```diff
@@ -0,0 +1,14 @@
+//import { lang } from './components/Shared/translations'
```
EFF, line 1: Well this is not necessary right ?

laurentlp: That's right, I removed it!

PR #485, opened 2022-05-20
chore(studio): remove unused stuff

EFF:


packages/studio/server/src/studio/testing/utils.ts
```diff
@@ -51,22 +50,24 @@ export const buildScenarioFromEvents = (events: sdk.IO.StoredEvent[]): Scenario
...
 ): Promise<string | undefined> => {
-  try {
-    const rows = await database('web_user_map').where({ botId, visitorId })
+  logger.error('An error occurred while fetching a visitor mapping. Not implemented')
+  return undefined
+  // TODO fix this
+  // try {
+  //   // const rows = await database('web_user_map').where({ botId, visitorId })
 
-    if (rows?.length) {
```
samuelmasse, line 62: This whole things makes no sense to be here anyways. web_user_map is an abomination to keep backward compatibility so we're certainly not keeping it. Let's leave it like this for now.

samuelmasse (approved):
Since there is no description : removed ndu frontend related stuff and removed some unused deps from the studio server, as well as unsuned env vars.

LGTM

PR #487, opened 2022-05-24
chore(messaging-server): remove spinned env var

samuelmasse:
Since the messaging server does not get "spinned" by botpress anymore, I removed the dead code related to that setting

### DIFF
--- a/packages/messaging/server/src/base/streamer.ts
+++ b/packages/messaging/server/src/base/streamer.ts
@@ -70,7 +70,7 @@ export class Streamer {
     const webhooks = await this.webhooks.list(clientId)
 
     for (const webhook of webhooks) {
-      void this.send(process.env.SPINNED_URL || webhook.url, payload, {
+      void this.send(webhook.url, payload, {
         'x-bp-messaging-client-id': clientId,
         'x-bp-messaging-webhook-token': webhook.token
       })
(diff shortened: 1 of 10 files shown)


samuelmasse:
Note: the concept of "lazy" channel does not make sense anymore. Should remove that

packages/messaging/server/src/rewire.ts
```diff
@@ -5,11 +5,7 @@ if (process.env.TS_NODE_DEV) {
   const originalRequire = Module.prototype.require as (id: string) => any
 
   const rewire = function (this: (id: string) => any, mod: string) {
-    if (
-      mod.startsWith('@botpress/messaging') &&
-      // we don't want to rewire the legacy channels because we get them from npm
-      mod !== '@botpress/messaging-channels-legacy'
-    ) {
+    if (mod.startsWith('@botpress/messaging')) {
```
laurentlp, line 8: So should we remove the legacy channel dependency too?

samuelmasse: Already removed that

packages/messaging/server/src/base/streamer.ts
```diff
@@ -78,21 +78,21 @@ export class Streamer {
   }
 
   public async send(url: string, data?: any, headers?: { [name: string]: string }) {
-    const config: AxiosRequestConfig<typeof data> = {}
-
-    if (headers) {
-      config.headers = headers
-    }
+    try {
+      const config: AxiosRequestConfig<typeof data> = {}
```
samuelmasse, line 83: Added after approval. Basically I moved the try so that it's over the entire function. Otherwise there is a small risk of unhandled rejection

PR #490, opened 2022-05-25
chore(messaging-channels): move content typings to own package

samuelmasse:
Creates a @boptress/messaging-content package that contains type definitions for our content types

Closes DEV-2669


packages/messaging/content/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "@botpress/messaging-content",
```
laurentlp, line 2: Why not `@botpress/messaging-content-types` or `@botpress/messaging-types` ? If find messaging-content to be confusing.

PR #502, opened 2022-05-30
chore(studio): remove testing from front-end code

laurentlp:
This PR removes the testing "module" from the frontend code since this will need to be rethought entirely.

Closes DEV-2681


samuelmasse (approved):
LGTM

PR #503, opened 2022-06-09
chore(tracing): added tracing to messaging server

michaelmass:
This adds an Opentelemetry tracing library to the messaging server. It can be configured by using the standard [environment variables](https://opentelemetry.io/docs/reference/specification/sdk-environment-variables).

The trace initialization MUST be done before importing any other package since it instruments known libraries like express, knex, pg. This also requires the trace initialization to be synced.

The important environment variables for configuring tracing are:

- TRACING_ENABLED `bool` Enables the tracer
- TRACING_DEBUG `bool` Adds debug information about the tracing configuration
- OTEL_SERVICE_NAME `string` Sets the service name given to a trace
- OTEL_EXPORTER_JAEGER_ENDPOINT `url` Sets the Jaeger collector endpoint

This is an example of what the tracing library allows:

image


packages/server/src/index.ts
```diff
@@ -1,3 +1,6 @@
+import { trace } from './tracing'
```
laurentlp, line 1: Can you add a eslint exception there

laurentlp (approved):
Small comment otherwise LGTM!

PR #512, opened 2022-07-28
chore(server): increase max payload size to 150kb

laurentlp:
Self-explanatory!

Note that we don't care if the max payload size is 100kb or 150kb as long as it's a reasonable size for text payloads. 150kb is also the max payload size of MS Teams.


sebburon (approved):
LTGM will better match Microsoft limit

PR #521, opened 2022-08-09
feat(webchat): allow to react on message click

laurentlp:
This PR allows the developer to react to message click on the webchat. When a user clicks on any message, an event (`messageClicked`) is sent to the parent windows (the HTML page that embeds the webchat) that contains the message details.


EFF (approved):
yess

PR #522, opened 2022-08-24
feat(webchat): allow users to subscribe to different webchat events

EFF:
**Done:** 
* converted inject.js to typescript ==> note that I didn't change the code much, I didn't want to go for a full re-implementation
* implemented event filtering and handling logic in the already existing window message listener in inject.ts ==> this not only improves the user experience it also prevent the user to create multiple window event listeners
* converted event types to types that are properly categorized
* fixed some typings along the way


packages/inject/src/inject.ts
```diff
@@ -0,0 +1,213 @@
...
+}
+
+function _injectDOMElement(
+  tagName: string,
+  selector: string,
+  options: { [key: string]: string } = {}
+): HTMLElement | void {
+  const element = document.createElement(tagName)
+  // @ts-ignore
+  Object.entries(options).forEach(([attrName, attrValue]) => (element[attrName] = attrValue))
```
laurentlp, line 37: **Suggestion:** if you ever think this adds value or could impress someone :sweat_smile: :

```ts
function _injectDOMElement<T extends keyof HTMLElementTagNameMap>(
  tagName: T,
  selector: string,
  options: {
    [key in keyof Partial<HTMLElementTagNameMap[T]>]?: HTMLElementTagNameMap[T][Extract<
      keyof HTMLElementTagNameMap[T],
      string
    >]
  } = {}
): HTMLElement | void {
  const element = document.createElement(tagName)

  for (const key in options) {
    element[key] = options[key]!
  }
// [...]
```

packages/inject/src/inject.ts
```diff
@@ -0,0 +1,213 @@
...
+  const parent = document.querySelector(selector)
+  if (!parent) {
+    console.error(`No element correspond to ${selector}`)
```
laurentlp, line 40: Shouldn't we throw here as this might be a "critical" error depending on the context?

packages/inject/src/inject.ts
```diff
@@ -0,0 +1,213 @@
...
+          types: []
+        }
+      }
+    },
+    set(target, prop: keyof WebchatRef, value) {
+      target[prop] = value
+      return true
+    }
+  }
+  return new Proxy<Partial<WebchatRef>>(target, chatRefHandler) as WebchatRef
```
laurentlp, line 99: Considering that you return a partial `WebchatRef` why cast it into a `WebchatRef`?

packages/inject/src/inject.ts
```diff
@@ -0,0 +1,213 @@
...
+  const chatRef = _getChatRef(chatId)
+  chatRef.postMessage({ action: 'mergeConfig', payload }, '*')
+}
+
+function onEvent(handler: WebchatEventHandler, topics: WebchatEventHandlerTopics = [], chatId?: string) {
+  if (typeof handler !== 'function') {
+    console.error('eventHandler is not a function, please provide a function')
+    return
+  }
+  if (!topics || typeof topics !== 'object' || !topics.length) {
```
laurentlp, line 136:
```suggestion
  if (!Array.isArray(topics)) {
```

packages/inject/src/inject.ts
```diff
@@ -0,0 +1,213 @@
...
+  if (CHAT_REFS[chatId]) {
+    Object.assign(CHAT_REFS[chatId], partialChatRef)
+  } else {
+    CHAT_REFS[chatId] = _makeChatRefProxy(chatId, partialChatRef)
+  }
+}
+
+/**
+ *
+ * @param {object} config Configuration object you want to apply to your webchat instance
```
laurentlp, line 153:
```suggestion
 * @param {Config} config Configuration object you want to apply to your webchat instance
```

packages/webchat/src/main.tsx
```diff
@@ -97,13 +96,9 @@ class Web extends React.Component<MainProps> {
     this.config = this.extractConfig()
     this.props.updateConfig!(this.config)
 
-    if (this.config.exposeStore) {
-      const storePath = this.config.chatId ? `${this.config.chatId}.webchat_store` : 'webchat_store'
-      set(window.parent, storePath, this.props.store)
-    }
-
+    // is this necessary ?
     if (this.config.containerWidth) {
```
laurentlp, line 100: I think some use it instead of adding custom CSS. We could deprecate it at some point tho!

packages/webchat/src/main.tsx
```diff
@@ -196,7 +192,6 @@ class Web extends React.Component<MainProps> {
           this.props.displayWidgetView ? this.props.showChat!() : this.props.hideChat!()
           trackWebchatState('toggle')
         } else if (type === 'message') {
-          trackMessage('sent')
```
laurentlp, line 199: Any particular reason why you removed this?

package.json
```diff
@@ -98,7 +98,7 @@
     "packages/*"
   ],
   "engines": {
-    "node": "16.13.2"
+    "node": "16.13"
```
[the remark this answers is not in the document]
EFF: don't you think because you and I and some other collaborator might already have another patch version of node 16.13 installed on their end. less anoying and between patches of node nothing should break ?.

If that is really annoying for you I can revert

packages/inject/src/inject.ts
```diff
@@ -0,0 +1,213 @@
...
+const DEFAULT_CHAT_ID = 'bp-web-widget'
+const DEFAULT_IFRAME_ID = 'bp-widget'
+
+const CHAT_REFS: { [chatId: string]: WebchatRef } = {}
+
+function _getContainerId(chatId?: string) {
+  return chatId ? `${chatId}-container` : DEFAULT_CHAT_ID
+}
+
+function _getIframeId(chatId: string) {
```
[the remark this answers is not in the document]
EFF: I can change it but in case you didn't know , return type is inferred by compiler. You get all the benefits of explicit typing with less code.

packages/inject/src/inject.ts
EFF: It seems to me that it's a lot playing around with ts, that makes the code less readable and less easy to maintain.

packages/inject/src/inject.ts
EFF: done

packages/inject/src/inject.ts
EFF: I don't return a partial, I return a proxy that behaves like a full WebchatRef

packages/inject/src/inject.ts
EFF: good catch this is a leftover of vanilla js

packages/webchat/src/main.tsx
EFF: because it's tracked in the sendMessage function as well ==> tracked twice

packages/inject/src/inject.ts
```diff
@@ -0,0 +1,213 @@
+import { Config, WebchatEvent, WebchatEventType } from '@botpress/webchat'
+
+require('./inject.css')
+
+interface WebchatRef {
+  postMessage: (message: any, targetOrigin: string) => void
+  eventListener: {
+    handler: (event: WebchatEvent) => void
```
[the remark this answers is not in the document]
EFF: done in other pr

package.json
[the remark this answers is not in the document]
laurentlp: Not annoying is just that I think we want to be using the same version of NodeJS as PKG. At the same time I agree that it should be okay to discard the path version here!

packages/inject/src/inject.ts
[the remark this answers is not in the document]
laurentlp: I get that, but explicitly specifying the return type will prevent you from let's say editing the function later on and returning undefined by accident!

packages/inject/src/inject.ts
laurentlp: It might impress some people tho ;)

laurentlp (approved):

laurentlp:
Note: The webchat test on firefox never completes on the CI but was tested locally and ran with success!

PR #524, opened 2022-08-29
feat(webchat): create conversation from iframe api

EFF:
This brings back the ability to create a conversation from the iframe api

### DIFF
--- a/packages/webchat/src/utils/webchatEvents.ts
+++ b/packages/webchat/src/utils/webchatEvents.ts
@@ -1,6 +1,9 @@
+import { cloneDeep } from 'lodash'
 import { WebchatEvent, WebchatEventType } from '../typings'
 
 export const postMessageToParent = (type: WebchatEventType, value: any, chatId: string) => {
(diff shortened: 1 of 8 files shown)


packages/webchat/src/utils/webchatEvents.ts
```diff
@@ -1,6 +1,9 @@
+import { cloneDeep } from 'lodash'
 import { WebchatEvent, WebchatEventType } from '../typings'
 
 export const postMessageToParent = (type: WebchatEventType, value: any, chatId: string) => {
+  //necessary because of potentially nested mobx proxy object isn't serializable
+  value = cloneDeep(value)
   const evt: WebchatEvent = { type, value, chatId }
```
laurentlp, line 7: **PP**: So we don't directly update the object passed in arg!

```suggestion
  const evt: WebchatEvent = { type, value: cloneDeep(value), chatId }
```

laurentlp (approved):
Small comment otherwise LGTM!

laurentlp (approved):

PR #535, opened 2022-09-09
chore(doc): add unofficial roadmap

laurentlp:


samuelmasse:
Also :

- Private bots (right now websocket are not protected by anything)
- Multichannel convo (I think this already works if you remove a unique constraint somewhere)
- A bit overkill but : multi user convo. Makes hitl implementation less of a hack if you can just add someone to the conversation
- For the sandbox you need a token mechanism. Otherwise anyone can setup a sandbox using your clientId (not a problem if all bots are public though). Right now it works by asking for a clientId and making a link to your endpoint
- Pulic converse (there's no way to provision API routes at the moment that don't require clientId + token. Could work with user tokens)
- Direct connect to the studio using websocket to allow bot developere to test their local bots with live or sandbox channels
- Automated testing of channels

There's probably more but that's what I can think of as of now. Next doc you could make a glossary. There's a lot of random words used all of the code (most famous one : conduit).

PR #537, opened 2022-09-13
feat(webchat): added bot info page to webchat

michaelmass:


spg (approved):

packages/webchat/src/typings.ts
```diff
@@ -321,6 +321,34 @@ export interface Config {
...
+   */
+  emailAddress?: string
+  /**
+   * Displays the bot's cover picture in the conversation page
+   */
+  coverPictureUrl?: string
+  /**
+   * Enables the bot's information page in the webchat
+   */
+  showBotInfoPage?: boolean
```
allardy, line 351: All the other configuration options relate to the bot's behavior, while all those properties are purely informational. What do you think about grouping them together? Ex:
```js
// or botInfo to be consistent
infoPage: {
  enabled: boolean
  website: string
...etc
}
```

PR #542, opened 2022-09-27
feat(events): support for "user fetched" event

AbrahamLopez10:
This will be used in the runtime to track "repeat users" of bots for Cloud Analytics.

### DIFF
--- a/packages/client/src/base.ts
+++ b/packages/client/src/base.ts
@@ -1,14 +1,17 @@
 import { uuid } from '@botpress/messaging-base'
 import axios, { AxiosInstance, AxiosRequestConfig } from 'axios'
 import cookie from 'cookie'
-import { ConversationStartedEvent, MessageFeedbackEvent, MessageNewEvent, UserNewEvent } from '.'
+import { ConversationStartedEvent, MessageFeedbackEvent, MessageNewEvent, UserCreatedEvent, UserFetchedEvent } from '.'
 import { MessagingClientAuth } from './auth'
 import { Emitter } from './emitter'
 import { Logger } from './logger'
 import { MessagingChannelOptions } from './options'
 
 export abstract class MessagingChannelBase extends Emitter<{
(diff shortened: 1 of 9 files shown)


samuelmasse (commented):
Something being fetched is an implementation detail. It looks like you're using the `fetch` function from the `UserService` as a metric to measure returning users. It's completely arbitrary that this would work at the moment, considering it's out of scope for the `fetch` function to be a tool used to measure returning users. Someone might create a new feature at some point that fetches user information at random intervals that would not correlate with user activity and that would pollute the metric.

Even if we don't consider that, it's still bizarre to treat the fetching of user info as being a returning user. A returning user is someone who *interacts* with the system some significant amount of time after their last interaction. Something being fetched is not an interaction, someone refreshing their screen, loading a website and ignoring the chatbot box, etc these are not returning users.

I would say in the case of chatbots a returning user is a user who sends a message about 1-2h after their last message. So the way of implementing this would be to create a new service (maybe something like `UserMetricsService`), you give `MessageService` as a dependency to that service and you listen on new message events. Then when a new message is created, you check the time of the last message for this conversation. If it's more than ~2h or something, you have a returning user and you can register this metric. To avoid having to fetch the last message every time a message is sent, you can cache the result and store it in an LRU cache that expires after ~1h30 so you don't have to check if the user is a returning user all the time. You can look at the `BillingService` to get an idea of how registering metrics based on messages being sent looks like. It's worth noting that this solution would work on all channels, not just the websocket.

Also this looks to me like a backend only feature, I don't see why the websocket should get notified when a user is fetched. The websocket has the ability to fetch the user info anyways, so why do we need to send it a confirmation that the user info was fetched on top of that? Consider the situation where a second websocket connection is created for the same user, you would get random user_fetched events on either connections.

AbrahamLopez10:
Cancelling PR for now as we'll follow a different approach to track new vs. repeat users for analytics.

PR #543, opened 2022-10-05
feat(telegram): allow receiving file/audio/video/image

davidvitora:
This PR adds support for receiving files (file/audio/video/image) with Telegram (channel v1.0.0 only).

Similar to: https://github.com/botpress/messaging/pull/480

### DIFF
--- a/packages/channels/src/telegram/api.ts
+++ b/packages/channels/src/telegram/api.ts
@@ -58,6 +58,22 @@ export class TelegramApi extends ChannelApi<TelegramService> {
   private async handleTelegrafMessage(scope: string, ctx: NarrowedContext<Context<Update>, Update.MessageUpdate>) {
     if ('text' in ctx.message) {
       await this.service.receive(scope, this.extractEndpoint(ctx), { type: 'text', text: ctx.message.text })
+    } else {
+      const type = Object.keys(ctx.message).find((i) => ['photo', 'document', 'video', 'audio'].includes(i))
+
+      if (!type) {
+        return
+      }
+
+      // @ts-ignore
+      // Get best photo
+      const item = type === 'photo' ? ctx.message.photo[ctx.message.photo.length - 1] : ctx.message[type]
+
+      await this.service.receive(scope, this.extractEndpoint(ctx), {
+        // Standardize types with messaging
+        type: type.replace('document', 'file').replace('photo', 'image'),
+        url: await ctx.telegram.getFileLink(item.file_id)
+      })
     }
   }
 
--- a/packages/channels/src/telegram/api.ts
+++ b/packages/channels/src/telegram/api.ts
@@ -1,6 +1,7 @@
 import { Response } from 'express'
+import _ from 'lodash'
 import { Context, NarrowedContext } from 'telegraf'
-import { Update } from 'telegraf/typings/core/types/typegram'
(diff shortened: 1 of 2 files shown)


packages/channels/src/telegram/api.ts
```diff
@@ -58,6 +58,22 @@ export class TelegramApi extends ChannelApi<TelegramService> {
...
+    } else {
+      const type = Object.keys(ctx.message).find((i) => ['photo', 'document', 'video', 'audio'].includes(i))
+
+      if (!type) {
+        return
+      }
+
+      // @ts-ignore
+      // Get best photo
+      const item = type === 'photo' ? ctx.message.photo[ctx.message.photo.length - 1] : ctx.message[type]
```
EFF, line 70: are media file always arrays ? if yes, do I understand correctly that you only send the last item ? is this really the intended behaviour ? 

if yes, then you should do the following `const item = ctx[type].pop()`.

but if you want my opinion, we should send all media elements in either one payload or multiple subsequent payloads

EFF (changes requested):
Please address comments and answer questions before we move forward with this

packages/channels/src/telegram/api.ts
```diff
@@ -58,6 +58,22 @@ export class TelegramApi extends ChannelApi<TelegramService> {
...
+
+      await this.service.receive(scope, this.extractEndpoint(ctx), {
+        // Standardize types with messaging
+        type: type.replace('document', 'file').replace('photo', 'image'),
```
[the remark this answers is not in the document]
davidvitora: I agree with you, i thought about it when creating this PR, and just wanted a confirmation from the engineering team, going to change this

packages/channels/src/telegram/api.ts
davidvitora: I searched for references regarding this, the array contains the same images but in different sizes, so we just care about the last one, since it has the best quality

EFF: then if photo is the only media with this behaviour, you should have a dedicated if statement to it.

pseudo : 
```ts
if(isText){
//  map to text payload
} else if (isImage) {
//  process photo payload 
} else if (isOtherMedia){
//  process file, video and audio payload 
} else {
// swallow payload because not supported ? 
// throw ? 
// downrender to text ? 
}
```

packages/channels/src/telegram/api.ts
```diff
@@ -58,6 +58,22 @@ export class TelegramApi extends ChannelApi<TelegramService> {
   private async handleTelegrafMessage(scope: string, ctx: NarrowedContext<Context<Update>, Update.MessageUpdate>) {
     if ('text' in ctx.message) {
       await this.service.receive(scope, this.extractEndpoint(ctx), { type: 'text', text: ctx.message.text })
+    } else {
+      const type = Object.keys(ctx.message).find((i) => ['photo', 'document', 'video', 'audio'].includes(i))
```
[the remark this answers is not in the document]
davidvitora: If you attach multiple files/images in just one message, Telegram is going to send each one as a separate message, so our implementation will work

packages/channels/src/telegram/api.ts
davidvitora: Done

EFF (approved):
👍 GG

PR #551, opened 2022-11-03
fix(messaging): fix inject error

allardy:
Pretty sure the error is related to the "isWebchatEvent" function... It was really too basic.
image

These event causes those errors:
image

Still need a way to test them correctly, since stratus uses the CDN, I didn't test "strictly", but logic looks fine.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -180,3 +180,4 @@ jobs:
       - name: Run tests
         run: |
           yarn test:chat --browser ${{ matrix.browser }}
+          yarn test:inject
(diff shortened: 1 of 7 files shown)


allardy:
https://linear.app/botpress/issue/STU-44/fix-annoying-error-coming-injectts

spg:
Adding tests for that function would be pretty easy. Should we do this now?

allardy:
Agreed that tests would be nice, but i'm not really used to DOM testing. Could be a good challenge for someone who implements new features in the future

allardy:
@spg Tests added.

I disabled the firefox test because it's broken and has been for a long time, 6h running time is abusive. Fixing the real error should be done in a sepatate pr

image
image

spg:
@allardy please add a ticket in Linear so we remember to fix the tests

spg (approved):

PR #553, opened 2022-11-10
chore(messaging): improve tests

allardy:
Tests are now completely end to end (previously they didn't test inject/webchat, which is what all end users uses.)

- Use `yarn serve` to start a server which responds with a webchat.html file with all the parameters pre-filled (this can also be used to test the inject script on the studio)
- Tests also validate that the correct lifecycle events are called (loaded, ready, message sent, message received)

Still some work to do to address flakiness and test more features, future PR. I'll evaluate the need for the waits and update tests later.


package.json
```diff
@@ -28,7 +28,8 @@
     "test:e2e:pg": "cross-env POSTGRESQL=true jest -c ./test/jest.e2e.config.ts",
     "test:sec:pg": "cross-env POSTGRESQL=true jest -c ./test/jest.security.config.ts",
     "test:mig:pg": "cross-env POSTGRESQL=true jest -c ./test/jest.migration.config.ts",
-    "test:chat": "TS_NODE_COMPILER_OPTIONS='{\"module\":\"esnext\",\"target\":\"es2017\",\"moduleResolution\":\"node\"}' node -r esm -r ts-node/register test/cypress/start.ts",
+    "test:chat:chrome": "start-test 'start' http://localhost:3100/status 'ts-node packages/inject/test/serve.ts' 3700 'cypress run --browser chrome'",
+    "test:chat:firefox": "start-test 'start' http://localhost:3100/status 'ts-node packages/inject/test/serve.ts' 3700 'cypress run --browser firefox'",
```
allardy, line 32: I know that's not ideal, but the tool start-test doesn't support command line arguments for now

EFF (approved):

PR #573, opened 2023-04-03
feat(slack): allow receiving file/audio/video/image

davidvitora:
This PR adds support for receiving files (file/audio/video/image) with Slack (channel v1.0.0 only).

Similar to https://github.com/botpress/messaging/pull/543

Also updated sqlite3 since install is failing on Mac M1


packages/channels/src/slack/api.ts
```diff
@@ -111,11 +111,23 @@ export class SlackApi extends ChannelApi<SlackService> {
...
-        scope,
-        { identity: '*', sender: message.user, thread: message.channel },
-        { type: 'text', text: message.text }
-      )
+      if ('files' in message && message.files?.length) {
+        for (const file of message.files) {
+          await this.service.receive(
+            scope,
+            { identity: '*', sender: message.user, thread: message.channel },
+            { type: this.mapMimeTypeToStandardType(file.mimetype), url: file.url_private, title: file.title }
```
EFF, line 119: is `url_private` reachable ?  or is it actually private

EFF:
Did you try to build and test a docker image ? Just to make sure sqlite works. It always been a tricky dep to update.

packages/channels/src/slack/api.ts
davidvitora: @EFF It is private, meaning that only people with access to the organization that the Bot is part of will have access to that image, but at the same time, only people from the org will have access to the bot anyways. The only way to have a public image, in this case, would be to have a service to re-upload those images to S3 (which we don't have to any channel, including those where the generated Image URL is temporary[telegram]). At the same time, it's totally possible to download this image in a hook using the botToken and the Slack API and upload it somewhere else.

EFF:
@davidvitora any updates ?

davidvitora:
@EFF Hey Eff, I believe that I answered your question regarding this PR. Is there anything else missing on my side?

EFF (approved):

michaelmass (approved):

PR #607, opened 2023-07-05
fix(webchat): convert plain URLs to links in markdown messages

AbrahamLopez10:


Tiltfile
```diff
@@ -7,7 +7,7 @@ config.set_enabled_resources([
 
 webchat_port = 3543
 inject_port = 8080
-messaging_port = 3100
+messaging_port = 3101
```
AbrahamLopez10, line 10: **note**: This is so messaging in this Tilt setup doesn't clash with messaging in Skynet by default, so it's easier to have this running alongside Skynet.

Tiltfile
```diff
@@ -37,7 +37,11 @@ local_resource(
 local_resource(
   name="webchat",
   serve_dir="packages/inject",
-  serve_cmd="yarn serve",
+  cmd="mkdir -p dist && yarn copy:shareable",
```
AbrahamLopez10, line 40: **note**: This fixes the webchat UI not being styled locally when serving from Tilt.

packages/components/src/utils.ts
```diff
@@ -1,13 +1,15 @@
 import { createIntl } from 'react-intl'
-import snarkdown from 'snarkdown'
```
AbrahamLopez10, line 2: **note**: snarkdown is quite outdated, so we're replacing it with MarkdownIt which supports converting plain URLs to links.

packages/components/src/utils.ts
```diff
@@ -1,13 +1,15 @@
...
 import { MessageConfig } from './typings'
 
-export const renderUnsafeHTML = (message: string = '', escaped: boolean): string => {
+export const markdownToHtml = (message: string = '', escaped: boolean): string => {
   if (escaped) {
     message = message.replace(/</g, '&lt;').replace(/>/g, '&gt;')
   }
 
-  const html = snarkdown(message)
+  const md = new MarkdownIt({ linkify: true })
```
AbrahamLopez10, line 10: **note**: The bug was that the webchat integration sends text messages with markdown=true always (by design, so that people can freely use markdown in the Studio), so to fix the issue we need to convert plain URLs in markdown to actual links.

michaelmass (approved):

PR #608, opened 2023-07-11
fix(webchat): correctly override color for all react-select dropdowns

AbrahamLopez10:


packages/inject/src/themes/shareable.css
```diff
@@ -1206,6 +1206,6 @@ Footer
   }
 }
 
-#react-select-2-listbox {
```
AbrahamLopez10, line 1209: **note**: This was only working for the first dropdown in the chatbot, as the number in the middle (the `2`) is incremented for each dropdown that appears.

spg (approved):

PR #615, opened 2023-08-08
feat(messaging): single tenant option

sebburon:


packages/channels/src/teams/stream.ts
```diff
@@ -32,9 +32,14 @@ export class TeamsStream extends ChannelStream<TeamsService, TeamsContext> {
...
         client_secret: config.appPassword,
+        tenant_id: config.tenantId,
         scope: 'https://api.botframework.com/.default'
       })
-      await axios.post('https://login.microsoftonline.com/botframework.com/oauth2/v2.0/token', params.toString())
+      if (config.tenantId) {
+        await axios.post(`https://login.microsoftonline.com/${config.tenantId}/oauth2/v2.0/token`, params.toString())
+      } else {
+        await axios.post('https://login.microsoftonline.com/botframework.com/oauth2/v2.0/token', params.toString())
+      }
```
EFF, line 42: That's literally the same logic, here's my suggestion: 
```suggestion
const tenant = config.tenantId ?? 'botframework.com'
await axios.post(`https://login.microsoftonline.com/${tenant}/oauth2/v2.0/token`, params.toString())
```

EFF (approved):
All good ✅

packages/channels/src/teams/stream.ts
sebburon: Even better thanks

spg (approved):

PR #636, opened 2023-09-19
fix: media limit issue

ptrckbp:
We had the wrong file in our previous PR. this updates the component theme-light.css file, which is the one included in all bots.

Related to https://github.com/botpress/messaging/pull/634


ptrckbp:
For some reason there are two theme-light.css files and we applied the fix to the wrong one in the previous PR!

AbrahamLopez10 (approved):

PR #666, opened 2024-03-20
fix: only send created event if there is data

EFF:


EFF:
Not necessary anymore

